PR #17, opened 2021-09-22
Tendermint v0.35.0-rc1

firelizzard18:
- Upgrades tendermint to `v0.35.0-rc1`, which supports follower nodes.
- Moves `package main` from `./main.go` to `./cmd/accumulated`.
- Refactors `./cmd/accumulated` to use subcommands instead of flags.
  - `accumulated -init -n Badlands` becomes `accumulated init -n badlands`
  - `accumulated -i 0` becomes `accumulated run -n 0`
  - This cleanly separates init vs run vs other concerns and makes it easy to add new features to the CLI.


blockchain/tendermint/tendermint.go
```diff
@@ -122,20 +126,23 @@ func Initialize(shardname string, index int, WorkingDir string) {
...
 
 	// Overwrite default config.
 	for i := 0; i < nValidators; i++ {
 		nodeDir := path.Join(WorkingDir, fmt.Sprintf("%s%d", defNodeName, i))
 		config.SetRoot(nodeDir)
-		config.LogLevel = "main:info,state:info,statesync:info,*:error"
+		config.Mode = "validator"
+		config.LogFormat = "plain"
+		config.LogLevel = "info"
+		// config.LogLevel = "main:info,state:info,statesync:info,*:error"
```
firelizzard18, line 145: `v0.35.0-rc1` switched to `zerolog`, and zerolog only supports a global log level.

blockchain/validator/Node.go
```diff
@@ -59,7 +59,7 @@ func (app *Node) Initialize(configFile string, workingDir string, key ed25519.Pr
 
 	laddr := viper.GetString("rpc.laddr")
 
-	rpcClient, _ := rpchttp.New(laddr, "/websocket")
+	rpcClient, _ := rpchttp.New(laddr)
```
firelizzard18, line 62: `rpchttp.New` no longer accepts a second parameter. There is another function that accepts options, but the path option defaults to `/websocket` so we don't need to use it.

router/jsonrpc_test.go
```diff
@@ -129,10 +130,9 @@ func _TestLoadOnRemote(t *testing.T) {
 
 func TestJsonRpcAnonToken(t *testing.T) {
 	//make a client, and also spin up the router grpc
-	dir, err := ioutil.TempDir("/tmp", "AccRouterTest-")
+	dir, err := ioutil.TempDir("", "AccRouterTest-")
```
firelizzard18, line 133: Leaving the first parameter blank tells `TempDir` to use the default temp dir.

bunfield (approved):

PR #18, opened 2021-09-23
Ac 250 basic router baseline

bunfield:


PaulSnow (approved):

PaulSnow:
Mostly accepted as is, because the commit is just huge.
Let's not collect so much outside develop if we can help it.

PR #23, opened 2021-09-25
Ac 288 restructure gen transaction and transactions

PaulSnow:
Hash the SigInfo and the transaction Hash, then hash them together.
Allows external merkle trees to feed into Accumulate.


firelizzard18:
LGTM

firelizzard18 (approved):

PR #37, opened 2021-09-27
StateDB consistency test

firelizzard18:
Adds a consistency test for StateDB


firelizzard18:
@PaulSnow This test demonstrates the issue I brought up. Also, it seems very odd to me that I have to call `sdb.GetPersistentEntry` in order to force the BPT's database batch to get written to disk.

In this case, `sdb.EnsureRootHash` can correct the issue, but:

1. It's a hack, and IMO loading state from disk really should restore everything into the expected state, without needing to do extra hashing.
2. It doesn't always work in production, for reasons that are not clear to me. Especially when I test on a cluster, in which case the hack works on node 0 (which perhaps incidentally is the one I send transactions to) but not on nodes 1 and 2.

PaulSnow (approved):

PR #40, opened 2021-09-27
Ac 313 bpt failed to flush root node when update is called

PaulSnow:
Add flush to Update in BPT.  Also added GetRootHash() to the manager.


firelizzard18 (approved):
LGTM

PR #41, opened 2021-09-28
Submit BPT DB batch when committing a block

firelizzard18:
Fixes AC-313. When commiting a block, submit the TXList of the BPT DB
manager, not just that of the MM DB manager.


firelizzard18:
@PaulSnow, #40 didn't completely solve the problem. The BPT DB TXList was still not getting flushed until the next call to `GetPersistentEntry`. With both of these changes, the state sync/block replay/consensus issue seems to be solved, barring a crash in between the two `EndBatch` calls.

PaulSnow (approved):

PR #46, opened 2021-09-29
CI validation

firelizzard18:
Run CI on all branches and PRs.

CI will fail on this PR because many tests are broken.


PaulSnow:
Shall we push anyway?

firelizzard18:
:+1:

PR #52, opened 2021-09-29
Improve load test experience

firelizzard18:
- Allow specifying the network name or number
- Add instructions to the README


Mayhallaj (approved):
Looks good

PR #53, opened 2021-09-29
Make README more dev-oriented

firelizzard18:
I pretty much just moved the bulk of the README into its own file. I didn't want the summary to be identical to the abstract, so I wrote something simple. I'm happy to expand the summary in the README if that needs to happen.


stolmanb (commented):
Do we want to mention Factom ?

PaulSnow (approved):

bunfield (approved):

PR #55, opened 2021-09-30
Fix JSON RPC tests

firelizzard18:
@bunfield ~~So far I've fixed a number of bugs (generally caused by me). It still isn't 100%, but this makes it less broken.~~ The tests succeed now.

### DIFF
--- a/router/jsonrpc_test.go
+++ b/router/jsonrpc_test.go
@@ -16,17 +16,25 @@ import (
 
 	"github.com/AccumulateNetwork/accumulated/internal/relay"
 	acctesting "github.com/AccumulateNetwork/accumulated/internal/testing"
+	"github.com/AccumulateNetwork/accumulated/networks"
 	"github.com/AccumulateNetwork/accumulated/types"
 	"github.com/AccumulateNetwork/accumulated/types/api"
 	"github.com/go-playground/validator/v10"
 )
 
-var testnet = flag.Int("testnet", 4, "TestNet to load test")
+var testnet = flag.String("testnet", "Localhost", "TestNet to load test")
 var loadWalletCount = flag.Int("loadtest-wallet-count", 100, "Number of wallets")
 var loadTxCount = flag.Int("loadtest-tx-count", 1000, "Number of transactions")
 
-func _TestLoadOnRemote(t *testing.T) {
-	txBouncer := relay.NewWithNetworks(*testnet)
+func TestLoadOnRemote(t *testing.T) {
+	if os.Getenv("CI") == "true" {
+		t.Skip("This test is not appropriate for CI")
+	}
+
+	txBouncer, err := relay.NewWithNetworks(networks.IndexOf(*testnet))
+	if err != nil {
+		t.Fatal(err)
+	}
 
 	_, privateKeySponsor, _ := ed25519.GenerateKey(nil)
 
@@ -88,8 +96,6 @@ func _TestLoadOnRemote(t *testing.T) {
 }
 
 func TestJsonRpcAnonToken(t *testing.T) {
-	t.Skip("broken test") // ToDo: Broken Test
-
 	//make a client, and also spin up the router grpc
 	dir, err := ioutil.TempDir("", "AccRouterTest-")
 	cfg := filepath.Join(dir, "Node0", "config", "config.toml")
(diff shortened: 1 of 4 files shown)


router/jsonrpc_test.go
```diff
@@ -16,17 +16,25 @@ import (
...
+var testnet = flag.String("testnet", "Localhost", "TestNet to load test")
 var loadWalletCount = flag.Int("loadtest-wallet-count", 100, "Number of wallets")
 var loadTxCount = flag.Int("loadtest-tx-count", 1000, "Number of transactions")
 
-func _TestLoadOnRemote(t *testing.T) {
-	txBouncer := relay.NewWithNetworks(*testnet)
+func TestLoadOnRemote(t *testing.T) {
+	if os.Getenv("CI") == "true" {
+		t.Skip("This test is not appropriate for CI")
+	}
```
firelizzard18, line 32: This test will be skipped when run from CI. This test either requires connecting to a remote IP or running a local node. Neither are appropriate for CI unit tests. If we want CI integration tests, we should set those up explicitly as integration tests.

router/jsonrpc_test.go
```diff
@@ -98,9 +104,10 @@ func TestJsonRpcAnonToken(t *testing.T) {
 	}
 	defer os.RemoveAll(dir)
 
-	_, pv, _ := startBVC(t, cfg, dir)
+	_, pv, node := startBVC(t, cfg, dir)
+	defer node.Stop()
 
-	txBouncer := relay.NewWithNetworks(*testnet)
+	txBouncer, err := relay.NewWithNetworks(networks.IndexOf("Badlands"))
```
firelizzard18, line 110: I made the original change, replacing a hard-coded number with `*testnet`, when I should not have.

router/jsonrpc_test.go
```diff
@@ -271,6 +281,7 @@ func TestJsonRpcAdi(t *testing.T) {
 	//now we can send in json rpc calls.
 	ret := jsonapi.createADI(context.Background(), jsonReq)
 
+	t.Skip("Needs acceptance criteria for create-adi response")
 	t.Fatal(ret)
```
firelizzard18, line 285: Under what conditions should this test fail?

router/utils_test.go
```diff
@@ -9,68 +9,98 @@ import (
...
+
+func boostrapBVC(t *testing.T, configfile string, workingdir string) {
+	t.Helper()
+
+	opts := initOptsForNetwork(t, "Badlands")
 	opts.WorkDir = workingdir
-	opts.Port = net.Port
-	opts.ListenIP = make([]string, len(net.Nodes))
-	for i, node := range net.Nodes {
-		opts.ListenIP[i] = node.IP
```
firelizzard18, line 41: I called the `node.Init` argument `Options`, but none of it is optional. All properties must be populated for everything to work right. I added `initOptsForNetwork`, which does basically the same thing as `accumulated init`.

router/utils_test.go
```diff
@@ -9,68 +9,98 @@ import (
...
-	//	size = 50000
-	//	wal_dir = ""
-	//
-
-	//viper.Set("mempool.keep-invalid-txs-in-cache, "false"
-	//viper.Set("mempool.max_txs_bytes", "1073741824")
-	viper.Set("mempool.max_batch_bytes", 1048576)
-	viper.Set("mempool.cache_size", 1048576)
-	viper.Set("mempool.size", 50000)
-	err := viper.WriteConfig()
```
firelizzard18, line 64: `node.Init` writes out the config file, so all we need to do is set the properties on `node.Options.Config` prior to calling `node.Init`.

router/utils_test.go
```diff
@@ -9,68 +9,98 @@ import (
...
+		Config:    config,
+		RemoteIP:  remoteIP,
+		ListenIP:  listenIP,
+	}
+}
+
+func boostrapBVC(t *testing.T, configfile string, workingdir string) {
+	t.Helper()
+
+	opts := initOptsForNetwork(t, "Badlands")
```
firelizzard18, line 72: Since we're only starting one node, we need to use `Badlands` and not `Localhost`. The latter will not work unless all three nodes are launched.

bunfield (approved):

PR #62, opened 2021-10-01
Don't hard-code TX relay

firelizzard18:
Make TX relay configurable instead of hard-coded to EastXeons.


firelizzard18:
@bunfield @PaulSnow Ready now. I made it a draft because I decided to add some extra validation to catch bad `--relay-to` entries early

bunfield (approved):

PR #85, opened 2021-10-07
Test node creation

firelizzard18:
Add tests for node creation.


bunfield (approved):

firelizzard18:
I updated the tests to use a free port instead of the hard coded port. Hopefully that will prevent issues with port reuse in tests.

PR #101, opened 2021-10-10
Test consistency

firelizzard18:
Create tests to ensure ABCI/chain validation/StateDB is consistent


bunfield (approved):

firelizzard18:
This doesn't actually work, so I'm closing it. The apparent test failure is due to an issue with taking a reference to an array without copying it. This test does not actually test consensus.

PR #105, opened 2021-10-11
Document URL specification

firelizzard18:
Document what we talked about so we don't trip up Jason with fun surprises

### DIFF
--- a/docs/URL.md
+++ b/docs/URL.md
@@ -1,0 +1,69 @@
+# Accumulate URLs
+
+A URL has a scheme, user info, authority, path, query, and fragment:
+
+```
+scheme://userinfo@authority/path?query#fragment
+```
+
+The scheme for Accumulate URLs must be `acc`. The `acc` scheme will be added if
+no scheme is specified, but specifying some other scheme will cause the
+transaction to be rejected.
+
+In general, the user info, query, and fragment components of the URL are not
+used.
+
+The authority of a URL may include a port number, but it will be ignored for
+most purposes. The hostname must not be empty. An empty authority or an
+authority that only includes a port number, e.g. `:80`, will be rejected.
+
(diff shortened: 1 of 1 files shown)


docs/URL.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+Thus, two URLs with the same authority have the same identity chain and routing,
+and two URLs with the same authority and path have the same resource chain,
+regardless of case.
+
+## Token Issuers
+
+The `ACME` token issuer is built into the protocol. In the future, other token
+issuers can be created. Every token issuer has a token URL, such as `acc://ACME`
```
PaulSnow, line 49: Need to specify if we have a token issuance chain or not.  Likely a token issuance chain would make our lives easier.

docs/URL.md
```diff
@@ -0,0 +1,64 @@
...
+or `acc://bobs/tokens`.
+
+## Anonymous Token Accounts
+
+The URL of an anonymous token account consists of the token URL appended to the
+public key hash. For example, for a key with a public key hash of `deadbeef`
+(not actually a valid SHA-256 hash), the anonymous token account URL for ACME
+tokens is `acc://deadbeef/ACME` and for `bobs/tokens` is
+`acc://deadbeef/bobs/tokens`. Any attempt to send tokens to an anonymous URL
+will be rejected if the URL does not match this pattern.
```
PaulSnow, line 59: deadbeef should be prefixed with the checksum of deadbeef/acme or deadbeef/bobs/tokens.

acc://kkkkkkkkdeadbeef/acme 

where kkkkkkkk is the first four bytes of the sha256 hash of "deadbeef/acme"

PaulSnow (changes requested):

PaulSnow (approved):

PR #106, opened 2021-10-12
Use new anonymous URL format

firelizzard18:
- Implements the new URL format: `acme-XXX` => `XXX/acme`
- Token URL is now `ACME` instead of `dc/ACME`


protocol/protocol.go
```diff
@@ -0,0 +1,97 @@
...
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/AccumulateNetwork/accumulated/internal/url"
+)
+
+const ACME = "ACME"
+
+// AcmeUrl returns `acc://ACME`
```
PaulSnow, line 16: Shouldn't this be `acc://acme` ?

PaulSnow (changes requested):

protocol/protocol.go
PaulSnow: Golden

PR #114, opened 2021-10-13
Ac 378 provide error reporting for batch dispatching

bunfield:
- refactored batching to collect error information from both single and batched transaction dispatches.
- cleaned up error reporting within the accumulator
- update jsonrpc to properly report errors received from check tx and memory pool
- updated jsonrpc tests to send in bogus transactions to make sure the errors are caught


internal/api/jsonrpc.go
```diff
@@ -316,16 +318,29 @@ func (api *API) sendTx(req *acmeapi.APIRequestRaw, payload []byte) *acmeapi.APID
...
 	if err != nil {
 		msg = []byte(fmt.Sprintf("{\"error\":\"%v\"}", err))
 		ret.Data = &msg
 		return ret
 	}
 
-	api.query.BatchSend()
+	stat := api.query.BatchSend()
+	resp := <-stat
+	close(stat)
```
firelizzard18, line 330: Generally channels should be closed by the sender, not the receiver. As long as the send goes through before the channel is closed, the value will get passed correctly. You can [create a buffered channel](https://play.golang.org/p/OXyqioOzDpb) or you can [block the sender](https://play.golang.org/p/ZheM_s_qoIA).

internal/api/jsonrpc.go
```diff
@@ -497,15 +512,32 @@ func (api *API) faucet(_ context.Context, params json.RawMessage) interface{} {
...
+	txInfo, err := api.query.BroadcastTx(gtx)
 	if err != nil {
 		return NewAccumulateError(err)
 	}
-	api.query.BatchSend()
+
+	stat := api.query.BatchSend()
+
+	res := <-stat
+	close(stat)
```
firelizzard18, line 523: Same comment about the sender closing instead of the receiver

PR #118, opened 2021-10-15
API/relay: Return deliver TX results

firelizzard18:
Updates the relay to block until both CheckTx and DeliverTx are complete. Updates the error/result reporting code to use `*ctypes.ResultBroadcastTxCommit` instead of `interface{}` to avoid (most) type checks.

Also pulls logger creation out of node.New so that the logger can be passed to the ABCI app.


PaulSnow (approved):

PaulSnow (changes requested):
didn't mean to approve the draft

PR #128, opened 2021-10-17
Run as a service

firelizzard18:
- Add support for installing and running as a mac/windows/linux service.
- Cleans up error handling in `cmd/accumulated`
- Custom logging, to feed logs to the service system

It should be noted that in Windows, stdout/stderr of a service is not captured. In Windows, if logs are not written to the service logger, they may as well not exist. Logs written to the service logger are visible in the event viewer, but everything else goes into a void.

### DIFF
--- a/internal/abci/accumulator.go
+++ b/internal/abci/accumulator.go
@@ -4,9 +4,9 @@ import (
 	"bytes"
 	"crypto/sha256"
 	_ "crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
-	"log"
 	"time"
 
 	"github.com/AccumulateNetwork/accumulated"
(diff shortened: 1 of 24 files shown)


internal/abci/accumulator.go
```diff
@@ -20,6 +20,7 @@ import (
 	abci "github.com/tendermint/tendermint/abci/types"
 	"github.com/tendermint/tendermint/crypto"
 	"github.com/tendermint/tendermint/crypto/encoding"
+	"github.com/tendermint/tendermint/libs/log"
```
PaulSnow, line 23: Why use tendermint's logging?  This has sometimes been an issue in the past for me, using basic functionality from a package that isn't providing that functionality as a core value.

PaulSnow (approved):

PR #149, opened 2021-10-20
Ac 435 ms hash ranges

PaulSnow:
FEAT: Add Range command to return a range of hashes within a chain.
The range is limited to 1/2 the Mark Frequency.  Since we are currently using a frequency of 256, the limit is 128 hashes per call.


firelizzard18 (commented):
Setting the chain ID as a state variable (struct field) seems really awkward. Could we not add it as a parameter to methods that need it?

PR #192, opened 2021-10-31
feat: disable TX subscription by default

firelizzard18:
Add a config option to enable TX subscription, and leave it disabled by default.


firelizzard18:
Please ignore the macOS build failure. I'm guessing I forgot to enable TX subscription somewhere (a lot of the tests rely on it). Since we're not using that in production, we can ignore the failure.

PaulSnow (approved):

PR #201, opened 2021-11-04
Added some tests to Simple Token Addresses

PaulSnow:
Added some validation rules to the Simple Token Address creation
Added a test
Enabled a test to old anonymous address generation

### DIFF
--- a/protocol/keys.go
+++ b/protocol/keys.go
@@ -4,9 +4,9 @@ import (
 	"bytes"
 )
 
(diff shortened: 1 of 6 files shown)


firelizzard18 (approved):
@PaulSnow I left a non-blocking suggestion.

On the methods that have unused parameters: those are called in code that is generated. Instead of adding complexity to the generator so it knows when parameters are unused, I gave everything the same signature, so the generator does not have to be as smart.

PR #208, opened 2021-11-05
Fix consensus error when you restart one node but not the other

firelizzard18:
See AC-489 and AC-490


firelizzard18:
@PaulSnow Would you look at this in the morning and see if you can figure out why we're getting duplicate transactions?

firelizzard18:
I renamed `Unused2` back to `Nonce` since A) we're still using it and B) Paul convinced me that we don't need/want a nonce for any of the signatures other than the first one.

bunfield (approved):

PR #223, opened 2021-11-12
merkle root value in query fix(AC-515)

ranjan008:
merkle hash in query response(AC-515)


firelizzard18:
In the future, please use branch names that include the Jira task, such as `AC-515` or `AC-515-mdroot`

firelizzard18 (approved):

PR #225, opened 2021-11-12
Rename transaction types

firelizzard18:


bunfield (approved):
Just left some observations, but otherwise approve.

types/transaction_types.go
```diff
@@ -1,169 +1,119 @@
...
-	TxTypeScratchEntry //per 256 bytes
-	TxTypeTokenCreate
-	TxTypeKeyUpdate //update keys on the keychain the identity
-	TxTypeMultisigTx
-	TxTypeStateQuery //sends a query to a chain and returns its state information
-	TxTypeCreateSigSpec
-	TxTypeCreateSigSpecGroup
-	TxTypeAddCredits
-	TxTypeUpdateKeyPage
+	TxUnknown = TransactionType(iota)
```
PaulSnow, line 18: Need more discussion about the needs for type changes.

If you want to freeze types, you should have explicit comments rather than use iota

PaulSnow (changes requested):

protocol/types_gen.go
```diff
@@ -141,28 +141,30 @@ func NewSigSpecGroup() *SigSpecGroup {
...
 }
 
-func (*AddCredits) GetType() types.TxType { return types.TxTypeAddCredits }
+func (*AddCredits) GetType() types.TransactionType { return types.TxAddCredits }
 
-func (*CreateSigSpec) GetType() types.TxType { return types.TxTypeCreateSigSpec }
+func (*CreateSigSpec) GetType() types.TransactionType { return types.TxCreateKeyPage }
 
-func (*CreateSigSpecGroup) GetType() types.TxType { return types.TxTypeCreateSigSpecGroup }
+func (*CreateSigSpecGroup) GetType() types.TransactionType { return types.TxCreateKeyBook }
```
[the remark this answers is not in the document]
firelizzard18: Yes

types/transaction_types.go
firelizzard18: Explicit comments and iota are orthogonal. I can add comments to all of the types and still use iota.

IMO, using iota is easier to read and less error-prone than adding an explicit numeric value to each entry.

internal/chain/_token_issuance.go
```diff
@@ -12,7 +12,7 @@ import (
 
 type TokenIssuance struct{}
 
-func (TokenIssuance) Type() types.TxType { return types.TxTypeTokenCreate }
+func (TokenIssuance) Type() types.TxType { return types.TxCreateToken }
```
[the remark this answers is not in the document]
firelizzard18: I see your point. I think I should just revert everything to `types.TxType...`, because what I did is really stylistic and an unnecessary change based purely on opinion (and not engineering). @PaulSnow, WDYT?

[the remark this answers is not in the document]
firelizzard18: I reverted this change, so everything is now `TxType...`

types/transaction_types.go
firelizzard18: All valid transaction types now have comments

firelizzard18: Iota is gone

PaulSnow (approved):

PR #227, opened 2021-11-13
Simplify chain validators

firelizzard18:
Simplifies the chain validators by merging CheckTx and DeliverTx into Validate. This is feasible since pending state changes and synthetic transactions are not sent until the transaction succeeds. Thus, CheckTx can use Validate and simply discard the pending changes.

This makes no functional changes.

### DIFF
--- a/internal/chain/create_key_book.go
+++ b/internal/chain/create_key_book.go
@@ -14,28 +14,28 @@ type CreateKeyBook struct{}
 
 func (CreateKeyBook) Type() types.TxType { return types.TxTypeCreateKeyBook }
 
-func checkCreateKeyBook(st *StateManager, tx *transactions.GenTransaction) ([]*protocol.SigSpec, *url.URL, error) {
+func (CreateKeyBook) Validate(st *StateManager, tx *transactions.GenTransaction) error {
 	if _, ok := st.Sponsor.(*state.AdiState); !ok {
-		return nil, nil, fmt.Errorf("invalid sponsor: want %v, got %v", types.ChainTypeAdi, st.Sponsor.Header().Type)
+		return fmt.Errorf("invalid sponsor: want %v, got %v", types.ChainTypeAdi, st.Sponsor.Header().Type)
 	}
 
 	body := new(protocol.CreateSigSpecGroup)
 	err := tx.As(body)
 	if err != nil {
-		return nil, nil, fmt.Errorf("invalid payload: %v", err)
+		return fmt.Errorf("invalid payload: %v", err)
 	}
 
 	if len(body.SigSpecs) == 0 {
-		return nil, nil, fmt.Errorf("cannot create empty sig spec group")
+		return fmt.Errorf("cannot create empty sig spec group")
 	}
 
 	sgUrl, err := url.Parse(body.Url)
 	if err != nil {
-		return nil, nil, fmt.Errorf("invalid target URL: %v", err)
+		return fmt.Errorf("invalid target URL: %v", err)
 	}
 
 	if !sgUrl.Identity().Equal(st.SponsorUrl) {
-		return nil, nil, fmt.Errorf("%q does not belong to %q", sgUrl, st.SponsorUrl)
+		return fmt.Errorf("%q does not belong to %q", sgUrl, st.SponsorUrl)
 	}
 
 	entries := make([]*protocol.SigSpec, len(body.SigSpecs))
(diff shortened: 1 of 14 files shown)


internal/chain/create_key_book.go
```diff
@@ -14,77 +14,63 @@ type CreateKeyBook struct{}
 
 func (CreateKeyBook) Type() types.TxType { return types.TxCreateKeyBook }
 
-func checkCreateKeyBook(st *StateManager, tx *transactions.GenTransaction) ([]*protocol.SigSpec, *url.URL, error) {
+func (CreateKeyBook) Validate(st *StateManager, tx *transactions.GenTransaction) error {
 	if _, ok := st.Sponsor.(*state.AdiState); !ok {
-		return nil, nil, fmt.Errorf("invalid sponsor: want %v, got %v", types.ChainTypeAdi, st.Sponsor.Header().Type)
+		return fmt.Errorf("invalid sponsor: want %v, got %v", types.ChainTypeAdi, st.Sponsor.Header().Type)
```
bunfield, line 19: Would this be a more meaningful error if it output the type name instead of number?  For example, types.ChainTypeAdi.Name() and st.sponsor.Header().Type.Name()

internal/chain/create_key_book.go
```diff
@@ -14,77 +14,63 @@ type CreateKeyBook struct{}
...
 	body := new(protocol.CreateSigSpecGroup)
 	err := tx.As(body)
 	if err != nil {
-		return nil, nil, fmt.Errorf("invalid payload: %v", err)
+		return fmt.Errorf("invalid payload: %v", err)
 	}
 
 	if len(body.SigSpecs) == 0 {
-		return nil, nil, fmt.Errorf("cannot create empty sig spec group")
+		return fmt.Errorf("cannot create empty sig spec group")
```
bunfield, line 29: Should we change name to Key Book here in the error.

internal/chain/create_key_book.go
```diff
@@ -14,77 +14,63 @@ type CreateKeyBook struct{}
...
+		return fmt.Errorf("%q does not belong to %q", sgUrl, st.SponsorUrl)
 	}
 
 	entries := make([]*protocol.SigSpec, len(body.SigSpecs))
 	for i, chainId := range body.SigSpecs {
 		entry := new(protocol.SigSpec)
 		err = st.LoadAs(chainId, entry)
 		if err != nil {
-			return nil, nil, fmt.Errorf("failed to fetch sig spec: %v", err)
+			return fmt.Errorf("failed to fetch sig spec: %v", err)
```
bunfield, line 46: "Failed to fetch Key Page"

internal/chain/create_token_account.go
```diff
@@ -14,60 +14,43 @@ type CreateTokenAccount struct{}
...
 	if !accountUrl.Identity().Equal(st.SponsorUrl) {
-		return nil, nil, nil, fmt.Errorf("%q cannot sponsor %q", st.SponsorUrl, accountUrl)
+		return fmt.Errorf("%q cannot sponsor %q", st.SponsorUrl, accountUrl)
 	}
 
-	if body.KeyBookUrl != "" {
-		keyBookUrl, err = url.Parse(body.KeyBookUrl)
+	account := state.NewTokenAccount(accountUrl.String(), tokenUrl.String())
+	if body.KeyBookUrl == "" {
+		account.SigSpecId = st.Sponsor.Header().SigSpecId
```
bunfield, line 41: Maybe not a fix for this PR, but while I'm thinking about it, sigspecid should be keybookid or keybookurl

internal/chain/withdraw_tokens.go
```diff
@@ -53,56 +44,41 @@ func checkTokenTx(st *StateManager, tx *transactions.GenTransaction) ([]*tokenSe
...
-	txAmt := big.NewInt(0)
-	for _, val := range body.To {
-		amt.Add(amt.AsBigInt(), txAmt.SetUint64(val.Amount))
-	}
-
-	if !account.CanDebitTokens(&amt.Int) {
-		return nil, nil, nil, nil, fmt.Errorf("insufficient balance")
+	total := types.Amount{}
+	for _, to := range body.To {
+		total.Add(total.AsBigInt(), new(big.Int).SetUint64(to.Amount))
```
bunfield, line 60: This isn't a problem we need to tackle now, so this is food for thought.  The "amount" is a uint64. If we support up to 18 decimals places of precision similar to the erc20 spec, then using unit64 will be extremely limiting for transactions.  There are uint128 implementations but those will greatly complicate marshaling in other languages.

internal/chain/create_key_book.go
firelizzard18: `%v` is smart and will call `.String()` on the value, so this will come out as `invalid sponsor: want ChainTypeAdi, got ChainTypeSigSpec` (for example).

internal/chain/create_key_book.go
firelizzard18: Yes, but I want to do that as part of [AC-528](https://accumulate.atlassian.net/browse/AC-528)

internal/chain/create_key_book.go
firelizzard18: Yes, but I want to do that as part of [AC-528](https://accumulate.atlassian.net/browse/AC-528)

internal/chain/create_token_account.go
firelizzard18: I intend to do this as part of [AC-533](https://accumulate.atlassian.net/browse/AC-533)

internal/chain/create_key_book.go
bunfield: ahhh...

bunfield (approved):

internal/chain/withdraw_tokens.go
firelizzard18: https://accumulate.atlassian.net/browse/AC-555

PR #228, opened 2021-11-13
sig spec group marshalling fix(AC-517)

ranjan008:
sig spec group marshalling fix(AC-517)

### DIFF
--- a/protocol/types_gen.go
+++ b/protocol/types_gen.go
@@ -1306,7 +1306,7 @@ func (v *SigSpecGroup) MarshalJSON() ([]byte, error) {
 	}
 	u.ChainHeader = v.ChainHeader
 	u.SigSpecs = chainSetToJSON(v.SigSpecs)
-	return json.Marshal(u)
+	return json.Marshal(&u)
 }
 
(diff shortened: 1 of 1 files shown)


protocol/types_gen.go
```diff
@@ -1307,7 +1307,7 @@ func (v *SigSpecGroup) MarshalJSON() ([]byte, error) {
 	}
 	u.ChainHeader = v.ChainHeader
 	u.SigSpecs = encoding.ChainSetToJSON(v.SigSpecs)
-	return json.Marshal(u)
+	return json.Marshal(&u)
```
firelizzard18, line 1310: Changes need to be made to `./internal/cmd/gentypes` - generated files should not be manually edited. CI is failing because this change desyncrhonizes `types.yml` and `types_gen.go`.

firelizzard18 (changes requested):

firelizzard18:
@ranjan008 After changing type generation, you need to run `go generate ./protocol` from the root (or `go generate` from the protocol directory) to regenerate everything that is affected by the change.

firelizzard18 (approved):

PR #229, opened 2021-11-14
fix for ignoring key page setting

bunfield:
- fixed cli command-line arguments for keybook/page when creating an adi.


firelizzard18 (approved):
This is a tiny change, so there's not much to comment on. We could add a comment, but command-line argument handling is pretty standard, so I'm not sure it needs a comment.

bunfield:
I added comments :)

PR #233, opened 2021-11-16
feat: check-only mode for API

firelizzard18:
For API v2, adds a `checkOnly` parameter to execute TX requests. If `checkOnly` is true, the transaction will be checked but not executed.

To verify, send an API v2 request with `checkOnly` set to `true`. The transaction should not be executed.


internal/api/v2/execute_test.go
```diff
@@ -11,9 +11,13 @@ import (
 	"testing"
 	"time"
 
+	. "github.com/AccumulateNetwork/accumulate/internal/api/v2"
+	mock_api "github.com/AccumulateNetwork/accumulate/internal/mock/api"
```
bunfield, line 15: snake case?

Sorry, I had to comment on something... :)

bunfield (approved):

internal/api/v2/execute_test.go
firelizzard18: The Go team's [recommendations on package names](https://go.dev/blog/package-names) state that package names should not include underscores or mixed case. However, `mock_api` is generated code for testing, so :shrug: I'm not going to worry about it.

PR #234, opened 2021-11-16
fixed get by chain id

bunfield:
- had to convert from hex to binary prior to passing into getByChainId()
- to test : cli get chain <chain id in hex>
- also note, the get by chain id is not supported on the testnet v0.2, so you should get back a jsonrpc error.

### DIFF
--- a/cmd/cli/cmd/get.go
+++ b/cmd/cli/cmd/get.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"log"
@@ -19,7 +20,17 @@ var getCmd = &cobra.Command{
 		switch args[0] {
 		case "chain":
 			if len(args) > 1 {
-				GetByChainId([]byte(args[1]))
+				//decode the inputs to binary
+				var chainId [32]byte
+				i, err := hex.Decode(chainId[:], []byte(args[1]))
+				if err != nil {
+					log.Fatalf("chain id input must be hex, %v", err)
+				}
+				//need to make sure the input is an appropriate length for a chain
+				if i != 32 {
+					log.Fatal("chain id must be 32 bytes in length")
+				}
+				GetByChainId(chainId[:])
 			} else {
 				fmt.Println("Usage:")
 				PrintGet()


firelizzard18:
I'm getting an error:

```
$ go run ./cmd/cli -d get chain aa4d6b62cad5f0bc4205bde7cc65e2a4217e103cb45c73989c47291a9f26cf83
{"chainId":"aa4d6b62cad5f0bc4205bde7cc65e2a4217e103cb45c73989c47291a9f26cf83","wait":false}
accumulated: https://devnet.accumulatenetwork.io/v1
--> {"jsonrpc":"2.0","method":"chain","params":{"chainId":"qk1rYsrV8LxCBb3nzGXipCF+EDy0XHOYnEcpGp8mz4M=","wait":false},"id":1121}
<-- {"jsonrpc":"2.0","error":{"code":-32802,"message":"Validation Error","data":"Key: 'APIRequestChainId.ChainId' Error:Field validation for 'ChainId' failed on the 'required' tag"},"id":1121}


2021/11/16 17:19:16 jsonrpc2.Error{Code:ErrorCode{-32802}, Message:"Validation Error", Data:"Key: 'APIRequestChainId.ChainId' Error:Field validation for 'ChainId' failed on the 'required' tag"}
exit status 1
```

firelizzard18:
You need to change line 75, `Client.Request`, to use `&params` instead of `params`. I'm not totally sure why that matters, but making that change fixes the error I got.

firelizzard18:
The query will still fail because the API doesn't work.
- `internal/api.Query.QueryByChainId` is missing `qu.Type = types.QueryTypeChainId`
- `case types.QueryTypeChainId` in `internal/chain.Executor.Query` needs to `chr.UnmarshalBinary(q.Content)` instead of `chr.UnmarshalBinary(chr.ChainId[:])`.

However, instead of fixing those issues, I would prefer for the CLI to use V2 for this functionality.

firelizzard18 (changes requested):
To fix the error I saw, change line 75, `Client.Request`, to use `&params` instead of `params`. I'm not totally sure why that matters, but making that change fixes the error I got.

PR #235, opened 2021-11-16
fix: node configuration

firelizzard18:
Updates `accumulated init` to use the correct default configuration.

To verify, run `accumulated init` and verify that `accumulate.toml` contains `prometheus-server = "http://18.119.26.7:9090"`. For example, `go run ./cmd/accumulated init -n Zion -r Zion`.


bunfield (approved):
I checked out the branch and confirmed the fix.

PR #239, opened 2021-11-17
chore: rename chain types

firelizzard18:
- Updates out-of-date chain types.
  - Anon token account => lite token account
  - Sig spec group => key book
  - Sig spec => key page
- Renames `ChainTypeAdi` to `ChainTypeIdentity` so that it matches `TxTypeCreateIdentity`
- Updates chain type JSON marshalling
  - `ChainTypeTokenAccount` now marshals to `"tokenAccount"` instead of `"ChainTypeTokenAccount"`

The only functional change is to the JSON marshalling of chain types.

### DIFF
--- a/internal/api/jsonrpc_test.go
+++ b/internal/api/jsonrpc_test.go
@@ -513,7 +513,7 @@ func TestQueryWrongType(t *testing.T) {
 	resp := japi.GetADI(context.Background(), req)
 	switch r := resp.(type) {
 	case jsonrpc2.Error:
-		require.Contains(t, r.Data, "want ChainTypeAdi, got ChainTypeAnonTokenAccount")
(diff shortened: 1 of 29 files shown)


internal/api/jsonrpc_test.go
```diff
@@ -513,7 +513,7 @@ func TestQueryWrongType(t *testing.T) {
 	resp := japi.GetADI(context.Background(), req)
 	switch r := resp.(type) {
 	case jsonrpc2.Error:
-		require.Contains(t, r.Data, "want ChainTypeAdi, got ChainTypeAnonTokenAccount")
+		require.Contains(t, r.Data, "want identity, got liteTokenAccount")
```
PaulSnow, line 516: want **ADI**, got **L**iteTokenAccount

PaulSnow (approved):

PR #242, opened 2021-11-17
AC-509 abci error code

ranjan008:
The pull request is made for the standardization of the response types for ABCI interfaces implementation. The return types are kept as tendermint internal response implementation as it provides internal methods to validate or access the response types.


internal/abci/abci.go
```diff
@@ -12,12 +12,12 @@
 package abci
 
 import (
+	"github.com/AccumulateNetwork/accumulate/types/api/transactions"
+	types2 "github.com/tendermint/tendermint/abci/types"
```
firelizzard18, line 16: These imports should be moved into the second block with the other github.com packages.

internal/abci/accumulator.go
```diff
@@ -165,24 +165,24 @@ func (app *Accumulator) InitChain(req abci.RequestInitChain) abci.ResponseInitCh
...
+		panic(fmt.Errorf("failed to execute genesis TX: %v", deliverRes.Info))
 	}
 
 	app.chain.EndBlock(EndBlockRequest{})
 
-	mdRoot, err := app.chain.Commit()
-	if err != nil {
+	commitRes := app.chain.Commit()
+	if commitRes.Data == nil {
 		panic(fmt.Errorf("failed to commit genesis TX: %v", err))
```
firelizzard18, line 182:
```suggestion
		panic(fmt.Errorf("failed to commit genesis TX: %v", commitRes.Info))
```

internal/abci/accumulator.go
```diff
@@ -249,9 +249,9 @@ func (app *Accumulator) CheckTx(req abci.RequestCheckTx) (rct abci.ResponseCheck
 	//create a default response
 	ret := abci.ResponseCheckTx{Code: 0, GasWanted: 1, Data: sub.ChainID, Log: "CheckTx"}
 
-	err = app.chain.CheckTx(sub)
+	checkRes := app.chain.CheckTx(sub)
 
-	if err != nil {
+	if !checkRes.IsOK() {
```
firelizzard18, line 254: The error code is not propagated to the caller. `ret.Code` is always set to `protocol.CodeUnknownError`, so the caller won't see the detailed error code.

In general, I think we just want to return `checkRes`. We want to retain the error logging, but I don't see a reason to create and populate a separate `var ret abci.ResponseCheckTx` given that `checkRes` has everything we want, I believe.

internal/abci/accumulator.go
```diff
@@ -308,6 +308,12 @@ func (app *Accumulator) DeliverTx(req abci.RequestDeliverTx) (rdt abci.ResponseD
 		return ret
 	}
 
+	r := new(protocol.TxResult)
+	err = json.Unmarshal(deliverRes.Data, r)
```
firelizzard18, line 312: I don't like this. I don't want to have to marshal and unmarshal the TX result, just to get around limitations imposed by reusing Tendermint's response types.

internal/encoding/errors.go
```diff
@@ -2,5 +2,56 @@ package encoding
...
+	CodeUnsupportedTxType           = 21
+	CodeImproperTxnRequest          = 22
+	CodeTxnCheckFailed              = 23
+	CodeDeliverTxError              = 24
+	CodeImproperAcceptedTxn         = 25
+	CodeImproperPendingTxn          = 26
+	CodeTxnStateError               = 27
+	CodeRecordTxnError              = 28
+	CodeSyntheticTxnError           = 29
+	CodeMarshallingError            = 30
```
firelizzard18, line 33: Add a comment to each error code describing under what conditions it is returned (using the standard Go comment pattern). For example, "CodeUnsupportedTxType is returned when the transaction type is not supported".

firelizzard18 (changes requested):
I'm not a big fan of reusing Tendermint's response types. IMO an interface that returns `error` is cleaner. I would prefer something along these lines:

```go
package protocol

type ErrorCode int

const (
    CodeOk           ErrorCode = 0
    CodeUnknownError ErrorCode = 1
    // ...
)

type Error struct {
    Code    ErrorCode
    Message string
}

var _ error = (*Error)(nil)

func (err *Error) Error() string {
    return err.Message
}
```

With `protocol.Error` above, `Accumulator` can check if the returned error is an instance of `protocol.Error`, and if it is, it can retrieve the error code.

firelizzard18 (changes requested):
I like this. The remaining issues are:

- ~~This removes calls to `recordTransactionError`. Those calls are needed to properly record that a transaction failed.~~
- Tests are failing
- Error codes are not propagated to the CheckTx/DeliverTx response
- ~~I think not-found is not propagated.~~
- ~~Is `internal/encoding/errors.go` used? If not, it should be removed.~~

internal/abci/accumulator.go
```diff
@@ -165,14 +165,14 @@ func (app *Accumulator) InitChain(req abci.RequestInitChain) abci.ResponseInitCh
 		Height:   -1,
 	})
 
-	err = app.chain.CheckTx(tx)
-	if err != nil {
-		panic(fmt.Errorf("failed to validate genesis TX: %v", err))
+	customErr := app.chain.CheckTx(tx)
+	if customErr != nil && customErr.Code != protocol.CodeOK {
```
firelizzard18, line 169: Unless I'm missing something, the executor never returns `CodeOK`, so we don't need this check.

internal/abci/accumulator.go
```diff
@@ -165,14 +165,14 @@ func (app *Accumulator) InitChain(req abci.RequestInitChain) abci.ResponseInitCh
...
+		panic(fmt.Errorf("failed to validate genesis TX: %v", customErr))
 	}
 
-	_, err = app.chain.DeliverTx(tx)
-	if err != nil {
-		panic(fmt.Errorf("failed to execute genesis TX: %v", err))
+	_, customErr = app.chain.DeliverTx(tx)
+	if customErr != nil && customErr.Code != protocol.CodeOK {
```
firelizzard18, line 174: Same comment about `CodeOK`

internal/abci/accumulator.go
```diff
@@ -249,9 +249,9 @@ func (app *Accumulator) CheckTx(req abci.RequestCheckTx) (rct abci.ResponseCheck
 	//create a default response
 	ret := abci.ResponseCheckTx{Code: 0, GasWanted: 1, Data: sub.ChainID, Log: "CheckTx"}
 
-	err = app.chain.CheckTx(sub)
+	customErr := app.chain.CheckTx(sub)
 
-	if err != nil {
+	if customErr != nil && customErr.Code != protocol.CodeOK {
```
firelizzard18, line 254: Same comment about `CodeOK`

internal/abci/accumulator.go
```diff
@@ -292,9 +292,9 @@ func (app *Accumulator) DeliverTx(req abci.RequestDeliverTx) (rdt abci.ResponseD
 	}
 
 	//run through the validation node
-	r, err := app.chain.DeliverTx(sub)
+	r, customErr := app.chain.DeliverTx(sub)
 
-	if err != nil {
+	if customErr != nil && customErr.Code != protocol.CodeOK {
```
firelizzard18, line 297: Same comment about `CodeOK`

firelizzard18 (changes requested):
Are the changes to `openrpc.json` intentional? I'm guessing you ran `go generate` or `go generate ./...`, which triggered the yaml2json job in `version.go`. If that is the case, please revert `openrpc.json`. No one other than me uses `openrpc.yml` so I should probably remove that and the corresponding go:generate job.

internal/abci/accumulator.go
ranjan008: This is if in case in future somebody returns CodeOk, then also it will work

internal/chain/executor.go
```diff
@@ -214,8 +214,10 @@ func (m *Executor) Query(q *query.Query) (k, v []byte, err *protocol.Error) {
 
 		var obj encoding.BinaryMarshaler
 		k, obj, err = m.queryByUrl(u)
-		if err != nil {
-			return nil, nil, &protocol.Error{Code: protocol.CodeTxnQueryError, Message: fmt.Sprintf("invalid URL in query %s", chr.Url)}
+		if err != nil && errors.Is(err, storage.ErrNotFound) {
+			return nil, nil, &protocol.Error{Code: protocol.CodeTxnQueryError, Message: err.Error()}
+		} else if err != nil {
+			return nil, nil, &protocol.Error{Code: protocol.CodeTxnQueryError, Message: err.Error()}
```
firelizzard18, line 220: If these two cases are identical by intent, there's no reason to have two cases

openrpc.json
```diff
@@ -68,736 +68,537 @@
             "$ref": "#/components/schemas/tokenAccount"
           }
         },
-        "errors": [
```
firelizzard18, line 71: Unless there is a specific reason why this changed, please revert these changes

protocol/errors.go
```diff
@@ -4,3 +4,57 @@ import "github.com/AccumulateNetwork/accumulate/internal/encoding"
...
+	CodeValidateTxnError ErrorCode = 21
+	//CodeInvalidTxnError is returned when txn doesn't contains proper data
+	CodeInvalidTxnError ErrorCode = 22
+	//CodeAddTxnError is returned when adding txn to state db fails
+	CodeAddTxnError ErrorCode = 23
+)
+
+type Error struct {
+	Code    ErrorCode
+	Message string
```
firelizzard18, line 53: After thinking about it for a while (and working with the code), I think this needs to be changed from `string` to `error`. The field name could change, but that's less important.

```suggestion
	Message error
```

protocol/errors.go
```diff
@@ -4,3 +4,57 @@ import "github.com/AccumulateNetwork/accumulate/internal/encoding"
...
+}
+
+var _ error = (*Error)(nil)
+
+func (err *Error) Error() string {
+	return err.Message
+}
```
firelizzard18, line 60: With `Message` changing to an `error`, we can implement `Unwrap() error`, which will allow `errors.Is` to work on protocol errors.

```suggestion
}

func (err *Error) Unwrap() error {
	return err.Message
}
```

internal/abci/accumulator.go
```diff
@@ -106,20 +106,20 @@ func (app *Accumulator) Query(reqQuery abci.RequestQuery) (resQuery abci.Respons
...
+	k, v, customErr := app.chain.Query(qu)
 	switch {
-	case err == nil:
-		// OK
+	case customErr == nil:
+		//Ok
 
-	case errors.Is(err, storage.ErrNotFound):
-		resQuery.Info = err.Error()
+	case strings.Contains(customErr.Error(), storage.ErrNotFound.Error()):
```
firelizzard18, line 114: With `protocol.Error.Unwrap` implemented and the return value from `queryByUrl` being passed through directly (or wrapped via `fmt.Errorf` with the `%w` format specifier), this can be reverted back to `errors.Is(err, storage.ErrNotFound)`. `strings.Contains` is not a great way to check for specific error values, given that error strings may change: https://stackoverflow.com/a/57613539/762175.

firelizzard18 (changes requested):

internal/abci/accumulator.go
```diff
@@ -106,20 +106,20 @@ func (app *Accumulator) Query(reqQuery abci.RequestQuery) (resQuery abci.Respons
...
 		return resQuery
 
 	default:
-		sentry.CaptureException(err)
-		app.logger.Debug("Query failed", "type", qu.Type.Name(), "error", err)
-		resQuery.Info = err.Error()
+		sentry.CaptureException(customErr)
+		app.logger.Debug("Query failed", "type", qu.Type.Name(), "error", customErr)
+		resQuery.Info = customErr.Error()
 		resQuery.Code = code.CodeTypeUnauthorized
```
firelizzard18, line 123:
```suggestion
		resQuery.Code = customErr.Code
```

internal/abci/accumulator.go
```diff
@@ -249,9 +249,9 @@ func (app *Accumulator) CheckTx(req abci.RequestCheckTx) (rct abci.ResponseCheck
 	//create a default response
 	ret := abci.ResponseCheckTx{Code: 0, GasWanted: 1, Data: sub.ChainID, Log: "CheckTx"}
 
-	err = app.chain.CheckTx(sub)
+	customErr := app.chain.CheckTx(sub)
 
-	if err != nil {
+	if customErr != nil {
 		u2 := sub.SigInfo.URL
```
firelizzard18, line 255: `ret.Code` needs to be set to `customErr.Code`

internal/abci/accumulator.go
```diff
@@ -292,9 +292,9 @@ func (app *Accumulator) DeliverTx(req abci.RequestDeliverTx) (rdt abci.ResponseD
 	}
 
 	//run through the validation node
-	r, err := app.chain.DeliverTx(sub)
+	r, customErr := app.chain.DeliverTx(sub)
 
-	if err != nil {
+	if customErr != nil {
 		u2 := sub.SigInfo.URL
```
firelizzard18, line 298: `ret.Code` needs to be set to `customErr.Code`

internal/chain/executor.go
```diff
@@ -319,7 +319,7 @@ func (m *Executor) check(tx *transactions.GenTransaction) (*StateManager, error)
 	default:
 		// The TX sponsor cannot be a transaction
 		// Token issue chains are not implemented
-		return nil, fmt.Errorf("invalid sponsor: chain type %v cannot sponsor transactions", sponsor.Header().Type)
+		return nil, fmt.Errorf("%v cannot sponsor transactions", sponsor.Header().Type)
```
firelizzard18, line 322: Why was this changed?

firelizzard18 (changes requested):

firelizzard18 (approved):

PR #246, opened 2021-11-18
test: fix compile error caused by refactor

firelizzard18:
This fixes a compile error. No functional changes.


firelizzard18:
I'm forcing this through since all it does is fix a compile failure in a new test.

PR #251, opened 2021-11-19
chore: Ac 519 tx output

bunfield:
- cleaned up human-readable output for tokenTx and syntheticTokenDeposit

output can be viewed by 

mishmash of receive and send token transactions
./cli tx history acc://7117c50f04f1254d56b704dc05298912deeb25dbc1d26ef6/ACME 0 10

receive tokens transaction example:
./cli tx get a5bea9be9885c5b71e8018f4391ea669d9b49bc41766e8b5f38e44b4c0abb720

send tokens transaction example:
./cli tx get 494aee6c6e7129450dd6086d720443592411673a22b05422325775ff77a1eb3c

### DIFF
--- a/cmd/cli/cmd/tx.go
+++ b/cmd/cli/cmd/tx.go
@@ -117,7 +117,7 @@ func GetTXHistory(accountUrl string, s string, e string) {
 		log.Fatal(err)
 	}
 
-	var res acmeapi.APIDataResponse
+	var res acmeapi.APIDataResponsePagination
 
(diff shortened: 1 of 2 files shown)


firelizzard18 (approved):
LGTM

firelizzard18:
There's a `go vet` failure, which is what's causing tests to fail:

```
cmd/cli/cmd/util.go:415:11: Sprintf format %s reads arg #2, but call has 1 arg
```

PR #253, opened 2021-11-22
state db as transactional (AC-459)

ranjan008:
converted state db as transactional @firelizzard18  at e2e_statedb_test.go one test is failing not sure if that requires block index validation change or need to check in the way I implemented the transactional part, if you can check once.


internal/chain/state.go
```diff
@@ -39,40 +39,33 @@ type storeState struct {
...
 	m := new(StateManager)
-	m.db = db
+	m.dbTx = dbTx
 	m.chains = map[[32]byte]state.Chain{}
 	m.stores = map[[32]byte]*storeState{}
 	m.writes = map[storage.Key][]byte{}
 	m.txHash = types.Bytes(tx.TransactionHash()).AsBytes32()
 	m.txType = tx.TransactionType()
 
-	// The genesis TX is special
```
firelizzard18, line 54: I think you need to merge with `develop`. Your changes are reverting some of the comments I added.

types/state/index.go
```diff
@@ -39,3 +39,31 @@ func (s *StateDB) GetIndex(index Index, chain []byte, key interface{}) ([]byte,
 	s.logInfo("GetIndex", "index", string(index), "chain", hex.EncodeToString(chain), "key", key, "computed", hex.EncodeToString(k[:]))
 	return s.Read(k)
 }
+
+func (tx *DBTransactional) Write(key storage.Key, value []byte) {
```
firelizzard18, line 43: Instead of adding `DBTransactional.Write` and `DBTransactional.WriteIndex` change the receiver of the old method. That ensures that no one can use Write or WriteIndex without a transaction, and it shows what changes if any there are to the method body.

types/state/stateTxn.go
```diff
@@ -0,0 +1,427 @@
...
+		sync:         s.sync,
+		updates:      s.updates,
+		transactions: s.transactions,
+		writes:       s.writes,
+		blockIndex:   s.blockIndex,
+	}
+	return dbTx
+}
+
+func (tx *DBTransactional) AddStateEntry(chainId *types.Bytes32, txHash *types.Bytes32, object *Object) {
```
firelizzard18, line 75: For now, change the receiver of the existing methods, do not create a new method. If you change the existing method, I can see the diff. If you add a new method, I can't easily compare that to the old method.

firelizzard18, line 75: For now, change the receiver of the existing methods, do not create a new method. If you change the existing method, I can see the diff. If you add a new method, I can't easily compare that to the old method.

types/state/stateTxn.go
```diff
@@ -0,0 +1,427 @@
...
+	"github.com/AccumulateNetwork/accumulate/smt/storage"
+	"github.com/AccumulateNetwork/accumulate/smt/storage/database"
+	"github.com/AccumulateNetwork/accumulate/types"
+	"github.com/tendermint/tendermint/libs/log"
+	"sort"
+	"sync"
+	"time"
+)
+
+type DBTransactional struct {
```
firelizzard18, line 22: I would prefer this be named `DBTransaction`, since it represents a database transaction.

types/state/stateTxn.go
```diff
@@ -0,0 +1,427 @@
...
+	debug        bool
+	bpt          *pmt.Manager //pbt is the global patricia trie for the application
+	blockIndex   int64        //Index of the current block
+	TimeBucket   float64
+	mutex        sync.Mutex
+	updates      map[types.Bytes32]*blockUpdates
+	writes       map[storage.Key][]byte
+	transactions transactionLists
+	sync         sync.WaitGroup
+	logger       log.Logger
```
firelizzard18, line 34: Instead of copying all the fields of `StateDB`, `DbTransactional` should have a `db *StateDB` field and should access other fields through that one.

firelizzard18 (changes requested):
Once you resolve the comment about changing the method receiver instead of creating a new method, `StateDB.updates`, `.transactions`, and `.writes` should no longer be used, and thus should be removed from `StateDB`.

types/state/stateTxn.go
```diff
@@ -0,0 +1,334 @@
...
+	"github.com/AccumulateNetwork/accumulate/smt/managed"
+	"github.com/AccumulateNetwork/accumulate/smt/storage"
+	"github.com/AccumulateNetwork/accumulate/smt/storage/database"
+	"github.com/AccumulateNetwork/accumulate/types"
+	"sort"
+	"sync"
+)
+
+type DBTransaction struct {
+	state *StateDB
```
firelizzard18, line 20: `updates`, `writes`, and `transactions` need to be removed from `StateDB` and added to `DBTransaction`.

types/state/stateTxn.go
```diff
@@ -0,0 +1,334 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to unmarshal state for %x", chainId)
+	}
+	//if verify {
+	//todo: generate and verify data to make sure the state matches what is in the patricia trie
+	//}
+	return ret, nil
+}
+
+func (tx *DBTransaction) AddTransaction(chainId *types.Bytes32, txId types.Bytes, txPending, txAccepted *Object) error {
```
firelizzard18, line 78: Instead of adding a copy of `AddTransaction`, change the receiver of the original.

types/state/stateTxn.go
```diff
@@ -0,0 +1,334 @@
...
+		}
+		//if we have valid data, store off the state
+		ret = currentState.stateData
+	}
+
+	return ret, nil
+}
+
+//AddSynthTx add the synthetic transaction which is mapped to the parent transaction
+func (tx *DBTransaction) AddSynthTx(parentTxId types.Bytes, synthTxId types.Bytes, synthTxObject *Object) {
```
firelizzard18, line 146: Instead of adding a copy of `AddSynthTx`, change the receiver of the original.

types/state/stateTxn.go
```diff
@@ -0,0 +1,334 @@
...
+
+	parentHash := parentTxId.AsBytes32()
+	if val, ok = tx.state.transactions.synthTxMap[parentHash]; !ok {
+		val = new([]transactionStateInfo)
+		tx.state.transactions.synthTxMap[parentHash] = val
+	}
+	*val = append(*val, transactionStateInfo{synthTxObject, nil, synthTxId})
+}
+
+func (tx *DBTransaction) Commit(blockHeight int64) ([]byte, int, error) {
```
firelizzard18, line 159: Instead of adding a copy of `Commit`, rename `WriteStates` and change its receiver.

types/state/stateTxn.go
```diff
@@ -0,0 +1,334 @@
...
+	tx.state.updates = make(map[types.Bytes32]*blockUpdates)
+
+	//return the state of the BPT for the state of the block
+	rh := types.Bytes(tx.state.RootHash()).AsBytes32()
+	tx.state.logInfo("WriteStates", "height", blockHeight, "hash", logging.AsHex(rh))
+	return tx.state.RootHash(), currentStateCount, nil
+	panic("TODO")
+}
+
+func (tx *DBTransaction) writeChainState(group *sync.WaitGroup, mutex *sync.Mutex, mm *managed.MerkleManager, chainId types.Bytes32) {
```
firelizzard18, line 246: Instead of adding a copy of `writeChainState`, change the receiver of the original.

types/state/stateTxn.go
```diff
@@ -0,0 +1,334 @@
...
+	//	if mdRoot == nil {
+	//		//shouldn't get here, but will reject if I do
+	//		panic(fmt.Sprintf("shouldn't get here on writeState() on chain id %X obtaining merkle state", chainId))
+	//	}
+	//	//todo:  Determine how we purge pending tx's after 2 weeks.
+	//	s.bpt.Bpt.Insert(chainId, *mdRoot)
+	//}
+}
+
+func (tx *DBTransaction) logInfo(msg string, keyVals ...interface{}) {
```
firelizzard18, line 301: Instead of adding `logInfo`, callers should call `tx.state.logInfo`

types/state/stateTxn.go
```diff
@@ -0,0 +1,334 @@
...
+func (tx *DBTransaction) Sync() {
+	tx.state.sync.Wait()
+}
+
+func (tx *DBTransaction) RootHash() []byte {
+	h := tx.state.bpt.Bpt.Root.Hash // Make a copy
+	return h[:]                     // Return a reference to the copy
+}
+
+func (tx *DBTransaction) writeBatches() {
```
firelizzard18, line 321: Instead of adding a copy of `writeBatches`, change the receiver of the original.

types/state/stateTxn.go
```diff
@@ -0,0 +1,334 @@
...
+
+func (tx *DBTransaction) logInfo(msg string, keyVals ...interface{}) {
+	if tx.state.logger != nil {
+		// TODO Maybe this should be Debug?
+		tx.state.logger.Info(msg, keyVals...)
+	}
+}
+
+func (tx *DBTransaction) GetDB() *database.Manager {
+	return tx.state.db
```
firelizzard18, line 309:
```suggestion
	return tx.state.GetDB()
```

types/state/stateTxn.go
```diff
@@ -0,0 +1,334 @@
...
+}
+
+func (tx *DBTransaction) Sync() {
+	tx.state.sync.Wait()
```
firelizzard18, line 313:
```suggestion
	tx.state.Sync()
```

types/state/stateTxn.go
```diff
@@ -0,0 +1,334 @@
...
+}
+
+func (tx *DBTransaction) writeBatches() {
+	defer tx.state.sync.Done()
+	tx.state.db.EndBatch()
+	tx.state.bpt.DBManager.EndBatch()
+}
+
+func (tx *DBTransaction) BlockIndex() int64 {
+	return tx.state.blockIndex
```
firelizzard18, line 328:
```suggestion
	return tx.state.BlockIndex()
```

types/state/stateTxn.go
```diff
@@ -0,0 +1,334 @@
...
+	return tx.state.db
+}
+
+func (tx *DBTransaction) Sync() {
+	tx.state.sync.Wait()
+}
+
+func (tx *DBTransaction) RootHash() []byte {
+	h := tx.state.bpt.Bpt.Root.Hash // Make a copy
+	return h[:]                     // Return a reference to the copy
```
firelizzard18, line 318:
```suggestion
	return tx.state.RootHash()
```

types/state/stateTxn.go
```diff
@@ -0,0 +1,334 @@
...
+	tx.state.bpt.DBManager.EndBatch()
+}
+
+func (tx *DBTransaction) BlockIndex() int64 {
+	return tx.state.blockIndex
+}
+
+func (tx *DBTransaction) EnsureRootHash() []byte {
+	tx.state.bpt.Bpt.EnsureRootHash()
+	return tx.state.RootHash()
```
firelizzard18, line 333:
```suggestion
	return tx.state.EnsureRootHash()
```

firelizzard18 (changes requested):

internal/abci/utils_test.go
```diff
@@ -121,7 +121,7 @@ func (n *fakeNode) NextHeight() int64 {
 }
 
 func (n *fakeNode) WriteStates() {
-	_, _, err := n.db.WriteStates(n.NextHeight())
+	_, _, err := n.db.Begin().Commit(n.NextHeight())
```
firelizzard18, line 124: I don't think `db.Begin().Commit(...)` serves any purpose. If tests still pass with `n.WriteStates()` removed, this method should be deleted. If that breaks tests, we should investigate why.

types/state/stateTxn.go
```diff
@@ -0,0 +1,78 @@
...
+type DBTransaction struct {
+	state        *StateDB
+	updates      map[types.Bytes32]*blockUpdates
+	writes       map[storage.Key][]byte
+	transactions transactionLists
+}
+
+func (tx *DBTransaction) WriteStates(blockHeight int64) ([]byte, int, error) {
+	panic("implement me")
+}
```
firelizzard18, line 20: Is this needed? My intent was to replace `StateDB.WriteStates` with `DBTransaction.Commit`, which you've done. So I think this can be deleted.

firelizzard18 (changes requested):

types/state/stateTxn.go
ranjan008: this method is required to make DBTransaction an implementation of DB interface otherwise it shows error.

firelizzard18 (approved):

PR #258, opened 2021-11-23
chore(chain): move queries to their own file

firelizzard18:
This is a simple copy-paste operation that moves `internal/chain.Executor.Query` and friends into their own file. Since querying and executing transactions are different operations, I think this will make it easier for developers to find what they're looking for.

No functional changes. To verify, check that the functions are identical and that queries still work (e.g. `cli adi get ...`).


bunfield:
looks fine...

bunfield (approved):

PR #261, opened 2021-11-24
Directory node stub

firelizzard18:
Implements a directory node stub, plus corresponding configuration and initialization changes.

To verify:
- `accumulated init -n DevNet.Zion` - works as usual, `accumulate.toml` specifies `directory = "DevNet.Directory"`
- `accumulated init -n DevNet.Directory` - four nodes, two validators and two followers, `accumulate.toml` specifies `type = "directory"`
- `accumulated init devnet` - creates a local devnet with a DN and one BVN

### DIFF
--- a/cmd/accumulated/cmd_init.go
+++ b/cmd/accumulated/cmd_init.go
@@ -6,13 +6,14 @@ import (
 	"net"
 	"net/url"
 	"os"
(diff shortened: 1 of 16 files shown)


firelizzard18:
Wrong branch name

PR #265, opened 2021-11-24
chore: Ac 611 improve error handling in cli

bunfield:
- removed all references to log.Fatal
- updated page and book get commands to be compatible with latest types_gen
- improved error handling to push all output and errors back up to the cobra command level so that data can be processed in a way that is suitable for unit testing. 
- No cli command changes in this PR
- When running the cli, it must be run with the -s https://devnet.accumulatenetwork.io


firelizzard18 (changes requested):
Something weird happened with type generation. Type gen first writes out the file, then calls `go fmt` on it. Maybe something is invalid so `go fmt` fails?

Otherwise, it LGTM.

bunfield:
Yeah, there is something weird going on. I merged with develop using my windows laptop and go fmt on windurs goes bonkers with embedded structs inside functions.   I'll fix.

firelizzard18 (approved):

PR #266, opened 2021-11-24
Enforce formating of imports

firelizzard18:
This is pretty pedantic, but it really annoys me when imports are organized randomly. This PR modifies the Lint workflow to enforce formatting of imports. It also adds `./internal/cmd/goimports`, which runs an import formatter against all Go files, and a go generate directive that runs that command. Thus, `go generate` will format imports across the entire project.

Alternatively, configuring your IDE to use `github.com/rinchsan/gosimports/cmd/gosimports` as a formatter will also work.


bunfield (approved):
looks good and like a handy thing to have.

PR #271, opened 2021-11-25
fix: regenerate abci mock

firelizzard18:
Regenerate ABCI mock to fix CI failures


firelizzard18:
I'm going to push this through to fix CI

PR #274, opened 2021-11-25
Add basic genesis app state to genesis doc

firelizzard18:
Updates node setup and ABCI chain initialization to preload the database. Node setup initializes a `memory.DB` and marshals it, saving it as the app state in the genesis document. ABCI chain initialization loads the app state (from the genesis doc) and pushes it into the database.

This will allow us to add arbitrary state to the node database during setup by directly manipulating the `memory.DB` and executing transactions. Right now, all it does is set "ChainID".

To verify, setup and launch a node, then stop it, then restart it. Verify that one of the startup messages is `Starting ABCI ... chain=<chainId> module=accumulate`. Previously, `chain=...` only appeared the first time the node was launched, as the chain ID was set by InitChain, which is only called the first time the node launches.

### DIFF
--- a/internal/abci/abci.go
+++ b/internal/abci/abci.go
@@ -35,6 +35,8 @@ type EndBlockRequest struct{}
 type Chain interface {
 	Query(*apiQuery.Query) (k, v []byte, err *protocol.Error)
 
+	InitChain(state []byte) error
+
 	BeginBlock(BeginBlockRequest)
 	CheckTx(*transactions.GenTransaction) *protocol.Error
 	DeliverTx(*transactions.GenTransaction) (*protocol.TxResult, *protocol.Error)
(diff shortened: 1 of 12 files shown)


internal/abci/abci.go
```diff
@@ -43,6 +45,9 @@ type Chain interface {
 }
 
 type State interface {
+	// ChainID returns the ID of the network's chain
+	ChainID() (string, error)
```
bunfield, line 49: I would recommend changing chainId to network id to avoid confusion between what tendermint thinks is a chain and what accumulate chains are.

bunfield (approved):

internal/abci/abci.go
firelizzard18: I changed it to `SubnetID`, since BVNs and the DN are subnetworks of the Accumulate network. WDYT?

bunfield: I think that is good.

PR #277, opened 2021-11-26
AC-586 query directory expand

sanderPostma:
https://accumulate.atlassian.net/browse/AC-586


internal/api/v2/types_gen.go
```diff
@@ -53,6 +53,11 @@ type QueryResponse struct {
 	Status      interface{}  `json:"status,omitempty" form:"status" query:"status" validate:"required"`
 }
 
+type DirectoryQueryResult struct {
```
sanderPostma, line 56: This was cloned from the v1 types_gen along with its MarshalBinary func because ExpandedEntries QueryResponse is a v2 type. The v1 version is still there because the v1 API is still using that.

internal/api/v2/types_gen.go
```diff
@@ -53,6 +53,11 @@ type QueryResponse struct {
...
+	Entries         []string         `json:"entries,omitempty" form:"entries" query:"entries" validate:"required"`
+	ExpandedEntries []*QueryResponse `json:"expandedEntries,omitempty" form:"expandedEntries" query:"expandedEntries" `
+}
```
sanderPostma, line 59: I also chose for a separate field name. Returning different structures for the same rest call can be painful to unmarshal in some languages.

internal/api/v2/types_gen.go
```diff
@@ -91,11 +96,18 @@ type TxResponse struct {
 }
 
 type UrlQuery struct {
-	Url   string `json:"url,omitempty" form:"url" query:"url" validate:"required,acc-url"`
+	Url string `json:"url,omitempty" form:"url" query:"url" validate:"required,acc-url"`
+}
+
+type QueryPagination struct {
 	Start uint64 `json:"start,omitempty" form:"start" query:"start"`
 	Count uint64 `json:"count,omitempty" form:"count" query:"count"`
```
sanderPostma, line 104: Since this is a separate interface now, maybe the validation for count can be set to required.

sanderPostma, line 104: Since this is a separate interface now, maybe the validation for count can be set to required.

internal/api/v2/types_gen.go
```diff
@@ -91,11 +96,18 @@ type TxResponse struct {
...
 }
 
+type QueryOptions struct {
+	ExpandChains bool `json:"expandChains,omitempty" form:"expandChains" query:"expandChains"`
+}
```
firelizzard18, line 109: All changes to these types needs to be done via `types.yml`. This file is generated and should not be edited directly. LMK if you need help with `types.yml`.

firelizzard18 (changes requested):

internal/api/v2/jrpc_query.go
```diff
@@ -17,13 +17,18 @@ func (m *JrpcMethods) Query(_ context.Context, params json.RawMessage) interface
...
 func (m *JrpcMethods) QueryDirectory(_ context.Context, params json.RawMessage) interface{} {
-	req := new(UrlQuery)
-	err := m.parse(params, req)
+	urlQuery := new(UrlQuery)
+	err := m.parse(params, urlQuery)
+	if err != nil {
+		return err
+	}
+	queryOptions := new(QueryOptions)
+	err = m.parse(params, queryOptions)
```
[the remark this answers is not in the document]
sanderPostma: My thoughts, jsonrpc is a relatively simple protocol with only methods & params and does not dictate entire schemas like OpenAPI3 does. This pushes typing concerns to the documentation and its implementation is entirely up to the implementors on both sides of the API. So in Golang within Accumulate something like _var queryDirParams struct { api.UrlQuery; api.QueryOptions }_ is fine, we have already the UrlQuery and QueryOptions structs within the project.
But users of other languages like Java/Kotlin/C# can use extends/implements, create separate classes or don't use classes/types at all and just load the params one by one.
I was trying to make sure the structs don't have params in there that are not supported causing clutter and potentially confusion. As we implement more and more query options specific to individual methods, we can't just keep appending al of them to UrlQuery.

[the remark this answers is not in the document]
sanderPostma: But the types.yml generator change things, I'll take another look.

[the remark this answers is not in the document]
firelizzard18: Sounds good, though I still want a comment.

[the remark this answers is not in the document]
sanderPostma: I decided it change it to be more clear, please take a look and let me know if you agree.

internal/api/v2/query_direct.go
```diff
@@ -235,3 +241,42 @@ func (q queryDirect) QueryTxHistory(s string, start, count int64) (*QueryMultiRe
...
+		return nil, fmt.Errorf("unknown response type: want chain, got %q", key)
+	}
+
+	obj, chain, err := unmarshalState(value)
+	if err != nil {
+		return nil, err
+	}
+
+	return packStateResponse(obj, chain)
+}
```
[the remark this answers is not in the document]
sanderPostma: Done

internal/api/v2/types_gen.go
sanderPostma: Done

internal/cmd/gentypes/main.go
```diff
@@ -146,6 +147,9 @@ func run(_ *cobra.Command, args []string) {
 				fmt.Fprintf(w, "\nstate.ChainHeader\n")
 			}
 		}
+		for _, e := range typ.Embeddings {
+			fmt.Fprintf(w, "\t\t%s\n", e)
+		}
```
firelizzard18, line 152:
```suggestion
		if len(typ.Embeddings) > 0 && !typ.NonBinary {
			check(fmt.Errorf("type %q: embedding is not supported for binary-marshalled types", typ.name))
		}
		for _, e := range typ.Embeddings {
			fmt.Fprintf(w, "\t\t%s\n", e)
		}
```

The binary marshal generator does not know about embedded fields, so until it does, we need to tell the user that is not supported.

protocol/types.yml
```diff
@@ -207,6 +207,14 @@ DirectoryQueryResult:
     type: slice
     slice:
       type: string
+    optional: true
+  - name: ExpandedEntries
+    type: slice
+    slice:
+      type: state.Object
+      pointer: true
+      marshal-as: self
```
firelizzard18, line 216: I had been assuming the type generator would trip on references to types from other packages. I'm happy to know that it works.

types/api/query/chain.go
```diff
@@ -15,6 +15,15 @@ type ResponseByUrl struct {
...
 }
 
+type RequestDirectory struct {
+	RequestByUrl
+	ExpandChains types.Bool
+}
+
+type ResponseDirectory struct {
+	state.Object
+}
```
firelizzard18, line 25: This type is not used anywhere, so it should be removed. I don't know why `ResponseByUrl` exists, we should delete that too (not necessarily in this PR).

firelizzard18 (changes requested):
Looks good, the only comments are minor.

types/api/query/chain.go
sanderPostma: Done (in next commit)
I also deleted ResponseByUrl since my cursor was already there...

internal/cmd/gentypes/main.go
sanderPostma: Done (in next commit)

Looking at this code I am also inclined to refactor like this:
```
printEmbeddings(typ, w) 
printFields(typ, w) 
fmt.Fprintf(w, "}\n\n") 
```
But that will give a much larger PR and gentypes is not really in scope of this ticket to begin with.

firelizzard18: The type generator could definitely use some love

firelizzard18 (approved):

PaulSnow:
Note you need to resync or rebase with develop

PR #278, opened 2021-11-26
AC-620 jsonrpc error handling

sanderPostma:
See https://accumulate.atlassian.net/browse/AC-620


internal/api/v2/jrpc.go
```diff
@@ -80,6 +80,10 @@ func NewJrpc(opts JrpcOptions) (*JrpcMethods, error) {
 		return nil, errors.New("local node specified but no client provided")
 	}
 
+	if opts.Config.DebugJSONRPC {
+		jsonrpc2.DebugMethodFunc = true
+	}
```
firelizzard18, line 85: Since this is global, maybe it should be located in `Node.Start` (`internal/node/node.go`) or `Program.Start` (`cmd/accumulated/program.go`)?

firelizzard18 (approved):
LGTM, with one minor suggestion. Tests need to be updated to fix compile errors.

PR #287, opened 2021-11-27
Correct synthetic transaction production

firelizzard18:
This PR correctly implements synthetic production, according to @PaulSnow and @bunfield's comments. It works like this:

- Block 1
  - Deliver
    - Transactions are processed and produce synthetic transactions
  - Commit
    - Synthetic transactions are written to the staging bucket, added to the BPT (hashed), and added to the anchor chain (hashed)
- Block 2
  - BeginBlock
    - The leader scans the anchor chain, generates signatures for the previous block's synthetic transactions, and sends a SyntheticSignTransactions transaction with the signatures to its own BVN
- Block 3
  - Deliver
    - The synthetic transaction signatures are validated
  - Commit
    - The signatures are written to the signatures bucket and added to the BPT (hashed)
- Block 4
  - BeginBlock
    - Pending signed synthetic transactions are batched and transmitted to BVNs
  - Commit
    - Signatures for sent transactions are removed from the signatures bucket
- Block 5 (all BVNs)
  - Synthetic transactions are processed

### DIFF
--- a/internal/abci/accumulator.go
+++ b/internal/abci/accumulator.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"runtime/debug"
 	"time"
 
 	"github.com/AccumulateNetwork/accumulate"
@@ -15,7 +16,6 @@ import (
 	"github.com/AccumulateNetwork/accumulate/protocol"
 	_ "github.com/AccumulateNetwork/accumulate/smt/pmt"
 	"github.com/AccumulateNetwork/accumulate/smt/storage"
-	"github.com/AccumulateNetwork/accumulate/types"
 	apiQuery "github.com/AccumulateNetwork/accumulate/types/api/query"
 	"github.com/AccumulateNetwork/accumulate/types/api/transactions"
 	"github.com/getsentry/sentry-go"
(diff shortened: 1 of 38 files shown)


firelizzard18:
This is ready for review. Tests are definitely still broken because of the changes to when and how synthetic transactions get settled.

internal/abci/accumulator.go
```diff
@@ -253,12 +267,33 @@ func (app *Accumulator) InitChain(req abci.RequestInitChain) abci.ResponseInitCh
...
 		Time:     req.Header.Time,
 	})
+	if err != nil {
+		app.fatal(err)
+		return ret
+	}
+
+	if false {
+		// Convert the synth TXN refs into events.
+		for _, syn := range r.SynthTxns {
```
bunfield, line 285: is this used for debugging?

bunfield:
The code looked fine.  The question I have is if Block 2 is scraping the anchor chain, is that is on the DN, or is that the root for the BVN of the previous block?

bunfield (approved):

internal/abci/accumulator.go
firelizzard18: I'll remove it. A while ago, I updated `DeliverTx` to produce Tendermint events whenever a synthetic transaction was produced. Due to the changes in this PR, that event production needed to move to `BeginBlock`. However, returning events here makes one of the tests panic in the middle of Tendermint. So I disabled it. But I don't think anyone is using those events, so I'll just remove it.

firelizzard18:
It's the root of the BVN of the previous block. At the end of the block, synthetic transaction IDs are appended to the anchor chain (BVN root) after anchors from records that changed. In the next block, the BVN scans its own anchor chain for those synthetic transaction IDs.

Paul suggested creating a separate chain just for the synthetic transaction IDs, since most consumers of the anchor chain won't care about synthetic transactions. I want to do that and some other work on chains, but not in this PR.

PR #293, opened 2021-11-29
Cli unit tests

bunfield:
- added unit tests for cli in accordance with test matrix
- currently only works with a BVN, not sure how well this will work with a DN in the mix since this creates a real node.  Need to find out if two networks can run inside of the same app. Earlier tests with multiple networks running in the same app were unsuccessful
- some tests are deferred until V2 api is in place, but this gives a basic framework for testing the CLI and having it map to the test matrix.


cmd/cli/cmd/root_test.go
```diff
@@ -0,0 +1,284 @@
...
+	opts, err := acctesting.NodeInitOptsForNetwork(acctesting.LocalBVN)
+	require.NoError(t, err)
+	opts.WorkDir = defaultWorkDir
+	opts.Port, err = net.GetFreePort()
+	require.NoError(t, err)
+	cfg := opts.Config[0]
+	cfg.Mempool.MaxBatchBytes = 1048576
+	cfg.Mempool.CacheSize = 1048576
+	cfg.Mempool.Size = 50000
+	cfg.Accumulate.API.EnableSubscribeTX = true
```
firelizzard18, line 104: For these tests, I think we should leave `EnableSubscribeTX` disabled. Test setup enables this because some tests use it. But we don't use it in production, so I think we should leave it disabled for test matrix tests.

firelizzard18 (commented):
The DN isn't used yet, so you can run just a BVN and everything should work.

It is possible to run multiple networks with multiple nodes in a single process (see `TestFaucetMultiNetwork` in `internal/node/e2e_test.go`. However, running multiple nodes in one process is not a current goal of Tendermint (https://github.com/tendermint/tendermint/issues/7076), so it's kind of messy. You certainly need to disable Prometheus - the second time any node sets up Prometheus, it will panic.

bunfield:
> The DN isn't used yet, so you can run just a BVN and everything should work.
> 
> It is possible to run multiple networks with multiple nodes in a single process (see `TestFaucetMultiNetwork` in `internal/node/e2e_test.go`. However, running multiple nodes in one process is not a current goal of Tendermint ([tendermint/tendermint#7076](https://github.com/tendermint/tendermint/issues/7076)), so it's kind of messy. You certainly need to disable Prometheus - the second time any node sets up Prometheus, it will panic.

We can do some interesting things if we use more than one network in a production environment, for sure, at least maybe simplify things as far as the DN goes.

firelizzard18 (approved):

PR #294, opened 2021-11-29
avoid duplicate entries in e2e_test

brohamgoham:
Screen Shot 2021-11-29 at 2 06 30 AM
used to make keyBook entry twice

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.16
 
 require (
 	github.com/AccumulateNetwork/jsonrpc2/v15 v15.0.0-20210802145948-43d2d974a106
+	github.com/BurntSushi/toml v0.4.1
 	github.com/DataDog/zstd v1.4.5 // indirect
 	github.com/Factom-Asset-Tokens/factom v0.0.0-20200222022020-d06cbcfe6ece
 	github.com/boltdb/bolt v1.3.1
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,6 @@ go 1.16
 
 require (
 	github.com/AccumulateNetwork/jsonrpc2/v15 v15.0.0-20210802145948-43d2d974a106
-	github.com/BurntSushi/toml v0.4.1
 	github.com/DataDog/zstd v1.4.5 // indirect
 	github.com/Factom-Asset-Tokens/factom v0.0.0-20200222022020-d06cbcfe6ece
 	github.com/boltdb/bolt v1.3.1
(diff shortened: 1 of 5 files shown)


internal/chain/synthetic_create_chain.go
```diff
@@ -91,10 +91,12 @@ func (SyntheticCreateChain) Validate(st *StateManager, tx *transactions.GenTrans
...
-			err = st.AddDirectoryEntry(u)
-			if err != nil {
-				return fmt.Errorf("failed to add ADI directory entry: %v", err)
-			}
+			if record.Header().SigSpecId != (types.Bytes32{}) {
+				err = st.AddDirectoryEntry(u)
+				if err != nil {
+					return fmt.Errorf("error adding directory entry: %v", err)
+					}
+				}
```
firelizzard18, line 99:
```suggestion
				}
			}
```

internal/chain/synthetic_create_chain.go
```diff
@@ -91,10 +91,12 @@ func (SyntheticCreateChain) Validate(st *StateManager, tx *transactions.GenTrans
 			}
 
 			// Update the ADI's directory index
-			err = st.AddDirectoryEntry(u)
-			if err != nil {
-				return fmt.Errorf("failed to add ADI directory entry: %v", err)
-			}
+			if record.Header().SigSpecId != (types.Bytes32{}) {
```
firelizzard18, line 94: This would result in the key book never getting added to the directory index, because the key book's SigSpecId is always zero. If you check `n.GetDirectory(adiUrl)` in `TestCreateADI`, with this change the directory is missing `acc://RoadRunner/foo-book`.

firelizzard18 (changes requested):
I think the essential issue is that SyntheticCreateChain can update records as well as create them, and a record should not be added to the directory index if it is being updated.

Tests in `internal/abci/e2e_test.go` need to be updated to verify that the directory index includes what we want it to.

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.16
 
 require (
 	github.com/AccumulateNetwork/jsonrpc2/v15 v15.0.0-20210802145948-43d2d974a106
+	github.com/BurntSushi/toml v0.4.1
```
brohamgoham, line 7: need to rebase

firelizzard18 (approved):

firelizzard18 (approved):

PR #297, opened 2021-11-30
Add missing transaction names

firelizzard18:


firelizzard18:
Dennis reviewed this when it was part of #287, so I'm going to force-merge.

PR #298, opened 2021-11-30
Fetch TX history across mark points

firelizzard18:
Update `StateDB.GetChainRange` to work if the specified range crosses a mark point.


firelizzard18:
Dennis reviewed this when it was part of #287, so I'm going to force-merge.

PR #300, opened 2021-11-30
Ac 510 rpc error codes

ranjan008:
rpc error codes for specific responses


internal/api/jsonrpc.go
```diff
@@ -357,13 +357,13 @@ func (api *API) getTokenAccountHistory(_ context.Context, params json.RawMessage
 
 	// validate URL
 	if err = api.validate.Struct(req); err != nil {
-		return validatorError(err)
+		return invalidTxnTypeError(err)
```
firelizzard18, line 360: Errors here indicate incorrect parameters, not an invalid transaction type

firelizzard18 (changes requested):
I like what you have here, but

1. This work needs to be done in API v2 instead of v1. In general, we are no longer working on API v1.
2. We need to map protocol error codes to JSON-RPC error codes. I suggest something long these lines:

```patch
diff --git a/internal/api/v2/error.go b/internal/api/v2/error.go
index fbfd80e..98a8bd2 100644
--- a/internal/api/v2/error.go
+++ b/internal/api/v2/error.go
@@ -3,6 +3,7 @@ package api
 import (
 	"errors"
 
+	"github.com/AccumulateNetwork/accumulate/protocol"
 	"github.com/AccumulateNetwork/accumulate/smt/storage"
 	"github.com/AccumulateNetwork/jsonrpc2/v15"
 	"github.com/getsentry/sentry-go"
@@ -30,6 +31,8 @@ const (
 	ErrCodeMetricsVectorEmpty
 )
 
+const ErrCodeProtocolBase = -33000
+
 var (
 	ErrInternal           = jsonrpc2.NewError(ErrCodeInternal, "Internal Error", "An internal error occured")
 	ErrCanceled           = jsonrpc2.NewError(ErrCodeCanceled, "Canceled", "The request was canceled")
@@ -49,6 +52,9 @@ func accumulateError(err error) jsonrpc2.Error {
 	if errors.Is(err, storage.ErrNotFound) {
 		return jsonrpc2.NewError(ErrCodeNotFound, "Accumulate Error", "Not Found")
 	}
+	if err, ok := err.(*protocol.Error); ok {
+		return jsonrpc2.NewError(jsonrpc2.ErrorCode(ErrCodeProtocolBase+err.Code), "Accumulate Error", err.Message)
+	}
 	return jsonrpc2.NewError(ErrCodeAccumulate, "Accumulate Error", err)
 }
 
diff --git a/internal/api/v2/jrpc_utils.go b/internal/api/v2/jrpc_utils.go
index 32234e2..59e7c09 100644
--- a/internal/api/v2/jrpc_utils.go
+++ b/internal/api/v2/jrpc_utils.go
@@ -1,6 +1,10 @@
 package api
 
-import "encoding/json"
+import (
+	"encoding/json"
+
+	"github.com/AccumulateNetwork/accumulate/protocol"
+)
 
 func (m *JrpcMethods) parse(params json.RawMessage, target interface{}, validateFields ...string) error {
 	err := json.Unmarshal(params, target)
diff --git a/internal/api/v2/query_direct.go b/internal/api/v2/query_direct.go
index ba13cae..621239d 100644
--- a/internal/api/v2/query_direct.go
+++ b/internal/api/v2/query_direct.go
@@ -57,18 +57,22 @@ func (q queryDirect) queryType(typ types.QueryType, content queryRequest) (strin
 		return "", nil, fmt.Errorf("failed to send request: %v", err)
 	}
 
-	switch {
-	case res.Response.Code == protocol.CodeNotFound:
+	if res.Response.Code == 0 {
+		return string(res.Response.Key), res.Response.Value, nil
+	}
+
+	if res.Response.Code == protocol.CodeNotFound {
 		return "", nil, storage.ErrNotFound
-	case res.Response.Log != "":
-		return "", nil, errors.New(res.Response.Log)
-	case res.Response.Info != "":
-		return "", nil, errors.New(res.Response.Info)
-	case res.Response.Code != 0:
-		return "", nil, fmt.Errorf("query failed with code %d", res.Response.Code)
 	}
 
-	return string(res.Response.Key), res.Response.Value, nil
+	protoErr := new(protocol.Error)
+	protoErr.Code = protocol.ErrorCode(res.Response.Code)
+	if res.Response.Info != "" {
+		protoErr.Message = errors.New(res.Response.Info)
+	} else {
+		protoErr.Message = errors.New(res.Response.Log)
+	}
+	return "", nil, protoErr
 }
 
 func (q queryDirect) QueryUrl(s string) (*QueryResponse, error) {
```

firelizzard18 (approved):

PR #301, opened 2021-11-30
Anon => Lite

firelizzard18:
No functional changes.


bunfield (commented):
Lines 85 and 93 of account.go has reference to anon

bunfield (approved):

PR #304, opened 2021-12-01
Attempt to get the CLI tests running in CI

firelizzard18:
By changing the block frequency to ~10Hz

### DIFF
--- a/cmd/cli/cmd/adi_test.go
+++ b/cmd/cli/cmd/adi_test.go
@@ -35,14 +35,14 @@ func testCase2_1(t *testing.T, tc *testCmd) {
 	testCase5_1(t, tc)
 
 	//need to wait a sec to make sure faucet tx settles
-	time.Sleep(10 * time.Second)
+	time.Sleep(2 * time.Second)
 
 	commandLine := fmt.Sprintf("adi create %s acc://RedWagon red1", liteAccounts[0])
 	_, err := tc.execute(t, commandLine)
 	require.NoError(t, err)
 
 	//need to wait 2 secs to make sure adi create settles
-	time.Sleep(10 * time.Second)
+	time.Sleep(2 * time.Second)
 
 	//if this doesn't fail, then adi is created
 	_, err = tc.execute(t, "adi directory acc://RedWagon")
@@ -115,7 +115,7 @@ func testCase2_6(t *testing.T, tc *testCmd) {
 
 	t.Log(r)
 	//need to wait 2 secs to make sure adi create settles
-	time.Sleep(10 * time.Second)
+	time.Sleep(2 * time.Second)
 
 	//if this doesn't fail, then adi is created
 	r, err = tc.execute(t, "adi directory acc://Redstone")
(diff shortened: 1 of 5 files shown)


cmd/cli/cmd/root_test.go
```diff
@@ -118,6 +118,7 @@ func NewTestBVNN(t *testing.T, defaultWorkDir string) (int, int) {
 	require.NoError(t, err)                           //
 	cfg.Accumulate.WebsiteEnabled = false             // Disable the website
 	cfg.Instrumentation.Prometheus = false            // Disable prometheus: https://github.com/tendermint/tendermint/issues/7076
+	cfg.Consensus.TimeoutCommit = time.Second / 10    // Increase block frequency
```
bunfield, line 121: I had thought the minimum was 1 second.

firelizzard18: I doubt 100ms is reasonable for any production system, but it doesn't break. I ran a quick test and the actual block frequency I get is more like 5Hz.

internal/testing/app_client.go
```diff
@@ -244,7 +251,10 @@ func (c *ABCIApplicationClient) execute(interval time.Duration) {
 		// Collect any queued up sends
 	collect:
 		select {
-		case sub := <-c.txCh:
+		case sub, ok := <-c.txCh:
+			if !ok {
+				return
+			}
```
firelizzard18, line 257: Make sure the loop exists here if `txCh` has been closed

internal/testing/app_client.go
```diff
@@ -211,22 +211,29 @@ func (c *ABCIApplicationClient) execute(interval time.Duration) {
...
+			close(sub.DidCheck)
+			close(sub.DidDeliver)
+			close(sub.DidCommit)
+			sub.Done = true
+		}
+
+		c.txActive = 0
+		c.txPend = map[[32]byte]bool{}
+		c.txCond.Broadcast()
+	}()
```
firelizzard18, line 230: Move the cleanup to a defer so it happens at all exit points

internal/testing/app_client.go
```diff
@@ -211,22 +211,29 @@ func (c *ABCIApplicationClient) execute(interval time.Duration) {
...
+			}
+			close(sub.DidCheck)
+			close(sub.DidDeliver)
+			close(sub.DidCommit)
+			sub.Done = true
+		}
+
+		c.txActive = 0
+		c.txPend = map[[32]byte]bool{}
+		c.txCond.Broadcast()
```
firelizzard18, line 229: Make sure no one gets stuck waiting

cmd/cli/cmd/adi_test.go
```diff
@@ -35,14 +35,14 @@ func testCase2_1(t *testing.T, tc *testCmd) {
 	testCase5_1(t, tc)
 
 	//need to wait a sec to make sure faucet tx settles
-	time.Sleep(10 * time.Second)
+	time.Sleep(2 * time.Second)
```
PaulSnow, line 38: Gosh, this sure looks like this should be a constant rather than a hard coded value.

PaulSnow (changes requested):

PaulSnow (approved):

PR #305, opened 2021-12-01
chore: update credit to dollar conversion rate

firelizzard18:
We need to be able to charge fees down to 1/100th of a cent. The simplest way to do this is to change the conversion rate to 1 dollar = 10,000 credits.

### DIFF
--- a/protocol/protocol.go
+++ b/protocol/protocol.go
@@ -24,10 +24,22 @@ func AcmeUrl() *url.URL {
 // AcmePrecision is the precision of ACME token amounts.
 const AcmePrecision = 1e8
 
-// CreditsPerDollar is the conversion rate from 1 unit of fiat currency to
-// credits. We call this 'dollars' because it's easier to write, and we are most
-// likely going to use USD idefinitely.
-const CreditsPerDollar = 1e2
+// FiatUnitsPerAcmeToken fixes the conversion between ACME tokens and fiat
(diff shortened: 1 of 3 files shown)


protocol/protocol.go
```diff
@@ -27,7 +27,9 @@ const AcmePrecision = 1e8
 // CreditsPerDollar is the conversion rate from 1 unit of fiat currency to
 // credits. We call this 'dollars' because it's easier to write, and we are most
 // likely going to use USD idefinitely.
-const CreditsPerDollar = 1e2
+//
+// 1 dollar = 10,000 credits
```
PaulSnow, line 31: Three choices:
Credit is 1/100 of a cent
Credit is 1 cent with two decimal places (1.00)
Credit is 1 dollar with four decimal places (1.0000)
Does not impact this code, but the path should be documented here.  My preference is a Penny with two decimal places because people can more easily understand paying 1.25 cents rather than .0125 dollars.

PaulSnow (approved):

protocol/protocol.go
firelizzard18: From a UX point of view, I prefer 1 cent with 2 decimal places. If we wanted, we could create a fake token issuer with a precision of 2.

firelizzard18: I reworded the comment and renamed things to be more clear

PaulSnow (approved):

PR #306, opened 2021-12-01
fix: reload state DB after loading genesis state

firelizzard18:
InitChain loads state from the genesis doc (AC-622, AC-632). However, the BPT root hash does not get updated. This PR adds a call to `StateDB.Load` to reload the BPT from the database. It also adds a sanity check.

To verify, create and launch a BVN node, wait for it to settle, stop it, and restart it. On `develop` this will fail. With this PR, it will not fail.

### DIFF
--- a/internal/chain/executor.go
+++ b/internal/chain/executor.go
@@ -17,6 +17,7 @@ import (
 	"github.com/AccumulateNetwork/accumulate/internal/logging"
 	"github.com/AccumulateNetwork/accumulate/protocol"
 	"github.com/AccumulateNetwork/accumulate/smt/common"
+	"github.com/AccumulateNetwork/accumulate/smt/pmt"
 	"github.com/AccumulateNetwork/accumulate/smt/storage"
 	"github.com/AccumulateNetwork/accumulate/smt/storage/memory"
 	"github.com/AccumulateNetwork/accumulate/types"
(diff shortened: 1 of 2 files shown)


internal/chain/executor.go
```diff
@@ -84,19 +85,47 @@ func NewExecutor(opts ExecutorOptions, executors ...TxExecutor) (*Executor, erro
...
+		hash = bpt.Root.Hash
+	case errors.Is(err, storage.ErrNotFound):
+		// OK
+	default:
+		return fmt.Errorf("failed to load BPT root hash from app state: %v", err)
+	}
+
+	// Dump the genesis state into the key-value store
 	dst := m.db.GetDB().DB
 	err = dst.EndBatch(src.Export())
```
PaulSnow, line 113: Export should create a map of the correct size when copying the DB, i.e.:

> // Export writes the database to a map
> func (m *DB) Export() map[storage.Key][]byte {
> 	m.mutex.Lock()
> 	defer m.mutex.Unlock()
> 
> 	ex := make(map[storage.Key][]byte,**len(m.entries)**)
> 	for k, v := range m.entries {
> 		ex[k] = v
> 	}
> 	return ex
> }

PaulSnow (changes requested):

PaulSnow (approved):

firelizzard18:
I'm going to merge this regardless of the failed builds, because maybe then the CLI test fix will pass CI

PR #307, opened 2021-12-01
Fix AC-664

firelizzard18:
Update typegen to not `omitempty` for certain fields


bunfield (approved):
no comment other than looks good.

PR #311, opened 2021-12-02
fix: go1.16 compat

firelizzard18:
Go 1.16 compatibility


firelizzard18:
I'm pushing this through to get continuous delivery working again

PR #312, opened 2021-12-02
FIX: added test to prevent writes on closed databases

PaulSnow:
Intended to avoid test failures in CI
Currently, I am still seeing some inconsistent failures even in my development environment.


firelizzard18 (approved):

smt/storage/badger/db.go
```diff
@@ -43,13 +46,21 @@ func (d *DB) InitDB(filepath string, logger storage.Logger) error {
 	if err != nil { // Panic if we can't open Badger
 		return err
 	}
+	d.DBOpen.Store(true)
 	return nil
 }
 
+func (d *DB) Ready() bool {
```
firelizzard18, line 53: Add a comment?

PR #314, opened 2021-12-02
Ac 544 state db readability part1

sanderPostma:
In this part I went through the code and renamed the vars which I could not understand by just reading and making them understandable without ctrl-b-ing into them. I know the Go style is keeping thing as short as possible but from a readability perspective this is not always helping, so I am trying to find a balance where it's as short as possible without losing the ability to read what it is for. (You also only need to write the var name in full the first time, all subsequent instances the IDE's autocomplete will write them for you.

I also broke up some functions to have a it shorter functions with more separation of concern. However micro-functions in Go are complicated because to constant need to handle err so I stay away from those.


firelizzard18 (approved):
LGTM

PR #315, opened 2021-12-03
Reduce chain manipulation complexity

firelizzard18:
- Create ChainManager to wrap chain manipulation logic
- Create a separate synthetic txns chain (remove from anchor chain)
- Treat the anchor chain like a normal chain


bunfield (approved):
looked ok to me.

PR #316, opened 2021-12-03
Ac 478 directory api pagination

sanderPostma:


types/helpers.go
```diff
@@ -317,25 +317,6 @@ func (s *String) UnmarshalBinary(data []byte) error {
 	return err
 }
 
-type Bool bool
```
sanderPostma, line 320: (I've added this in AC-586, but it's not consistent to use anymore)

firelizzard18: :+1:

firelizzard18 (approved):

bunfield (approved):

PR #319, opened 2021-12-03
Ac 616 create data account tx

bunfield:
- added data account state
- added managerChainUrl to CreateDataAccount (this may be more appropriate to move to ChainHeader in the future)
- added validator for the creation of the data account state
- added computation of entry hash (to be used in next incoming PR)


internal/chain/create_data_account.go
```diff
@@ -20,5 +21,57 @@ func (CreateDataAccount) Validate(st *StateManager, tx *transactions.GenTransact
...
+
+	//if we have a manger book URL, then we need to parse it.
+	var managerBookUrl string
+	if len(body.ManagerKeyBookUrl) != 0 {
+		u, err := url.Parse(body.ManagerKeyBookUrl)
+		if err != nil {
+			return fmt.Errorf("manager key book specified, but url is invalid: %v", err)
+		}
+		managerBookUrl = u.String()
+	}
```
firelizzard18, line 42: Is it OK for the manager key book to belong to a different ADI?

internal/chain/create_data_account.go
```diff
@@ -20,5 +21,57 @@ func (CreateDataAccount) Validate(st *StateManager, tx *transactions.GenTransact
...
+			return fmt.Errorf("invalid key book %q: %v", keyBookUrl, err)
+		}
+		copy(account.KeyBook[:], keyBookUrl.ResourceChain())
+	}
+
+	if managerBookUrl != "" {
+		c, err := st.LoadString(managerBookUrl)
+		if err != nil {
+			return fmt.Errorf("invalid key book %q: %v", managerBookUrl, err)
+		}
```
firelizzard18, line 68: If the manager is allowed to belong to another ADI, it might not be present on this BVN

internal/chain/state.go
```diff
@@ -401,6 +401,10 @@ func unmarshalRecord(obj *state.Object) (state.Chain, error) {
 		record = new(protocol.KeyPage)
 	case types.ChainTypeKeyBook:
 		record = new(protocol.KeyBook)
+	case types.ChainTypeDataAccount:
+		record = new(protocol.DataAccount)
+	case types.ChainTypeLiteDataAccount:
+		record = new(protocol.LiteDataAccount)
 	default:
 		return nil, fmt.Errorf("unrecognized chain type %v", header.Type)
 	}
```
firelizzard18, line 410: IIRC, this switch statement exists in at least three places: here, API v1, and API v2.

Maybe we should add `func NewChain(typ types.ChainType) (state.Chain, error)` to `package protocol`<sup>1</sup>? Then we could call that instead of duplicating the switch statement.

<sup>1</sup>We can't add it to `package types` because that would create a circular dependency.

types/state/DataAccount.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"bytes"
+	"fmt"
+
+	"github.com/AccumulateNetwork/accumulate/types"
+)
+
+type DataAccount struct {
+	ChainHeader
+	ManagerKeyBookUrl types.String `json:"managerKeyBookUrl"` //this probably should be moved to chain header
```
firelizzard18, line 12: Can we go ahead and put it in the chain header now? We're going to have to wipe the testnet anyways, so we might as well add it to the chain header now.

types/state/DataAccount.go
```diff
@@ -0,0 +1,72 @@
+package state
+
+import (
+	"bytes"
+	"fmt"
+
+	"github.com/AccumulateNetwork/accumulate/types"
+)
+
+type DataAccount struct {
```
firelizzard18, line 10: `DataAccount` is also defined in `protocol/types.yml`. That one should be removed if we're going to add it here.

types/state/DataAccount.go
bunfield: I almost did that, but didn't.  Sure I'll move it over.

types/state/DataAccount.go
bunfield: what is the preference, there is also state/types.yml.  There isn't anything special about the state that needs customization, so I can just remove this file and put it in state/types.yml if that is where things are headed.  The only specialization I suppose is the chain type id.

bunfield: I'm going to use the protocol version seems that is what the other ones are doing.

types/state/DataAccount.go
bunfield: do we want to maintain the manager keey book as a URL or chainID?  I prefer Url.

internal/chain/create_data_account.go
bunfield: I asked paul. if it was it would add a lot of complexity I would think.

firelizzard18: Yeah. For now, I'm inclined to require that the manager belong to the same ADI. We can always remove the restriction later.

types/state/DataAccount.go
firelizzard18: URL definitely. I think everyone prefers url and Paul seem almost surprised that it wasn't.

types/state/DataAccount.go
firelizzard18: I prefer generated types, because marshaling is boilerplate stuff, and with generated types we only have one thing to debug (the generator).

internal/chain/create_data_account.go
PaulSnow: Yes

internal/chain/create_data_account.go
PaulSnow: We will need to sign from a different ADI to put data in lite data accounts. (to pay the credits).
I believe it will not be so complex once we have the synthetic tx infrastructure in place.

internal/chain/state.go
PaulSnow: Can we do this in a subsequent pull request?

protocol/entry_hash.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"crypto/sha256"
+
+	"github.com/AccumulateNetwork/accumulate/smt/managed"
+)
+
+// ComputeEntryHash
+// returns the entry hash given external id's and data associated with an entry
+func ComputeEntryHash(root []byte, extIds [][]byte, data []byte) []byte {
```
PaulSnow, line 11: We discussed this as being simply:

`func ComputeEntryhash(data[][]byte) []byte {`

Don't we want to do this?

types/state/DataAccount.go
PaulSnow: Almost surprised....  Nothing surprises me (why does that work better with a haughty english accent?)

Yes URLs.  Because chainIDs only work within the context of an ADI (where all components must be on the same BVN now and forever).  If a ChainID isn't defined on a BVN, it becomes a search to find where it is defined.

PaulSnow (changes requested):

internal/chain/state.go
bunfield: Added jira task

internal/api/v2/unmarshal.go
```diff
@@ -49,6 +50,10 @@ func chainFromStateObj(obj *state.Object) (state.Chain, error) {
 		chain = new(protocol.KeyBook)
 	case types.ChainTypeTransaction:
 		chain = new(state.Transaction)
+	case types.ChainTypeDataAccount:
+		chain = new(protocol.DataAccount)
+	case types.ChainTypeLiteDataAccount:
+		chain = new(protocol.LiteDataAccount)
```
bunfield, line 56: I only added this here for consistency.  I added a new jira task to consolidate the switches

protocol/entry_hash.go
bunfield: yes, we do.  fixed it.

internal/chain/create_data_account.go
bunfield: i removed the check requiring the manager to be on the same adi.

internal/api/v2/unmarshal.go
firelizzard18: :+1:

firelizzard18 (changes requested):
Looks like this includes a binary that escaped .gitignore.

firelizzard18 (approved):

PaulSnow (approved):

PR #325, opened 2021-12-07
Fix synth anchor txn race condition

firelizzard18:
- Moves synth anchor txn creation to Commit
- Writes anchor txid to anchor chain head, instead of a separate kvdb entry
- See AC-690 for details on the race condition

To verify, run `rm -rf .nodes && go run ./cmd/accumulated init devnet -w .nodes -v 1 -f 0 --no-empty-blocks && go run ./cmd/accumulated run devnet -w .nodes` and verify that the DN receives a syntheticAnchor transaction. Previously, if the race condition was hit, you would see `Check failed error="failed to get txn ...: not found: no transaction defined for ..." type=syntheticSignTransactions` from the BVN and no syntheticAnchor from the DN.


bunfield (approved):
ran the code and got no errors. everything else looks ok to me

PR #327, opened 2021-12-08
feat(testing): enable socket listener for either Unix or TCP between tendermint && accumulate

brohamgoham:
feat(testing): enable socket listener for either Unix or TCP between tendermint && accumulate

### DIFF
--- a/internal/accumulated/run.go
+++ b/internal/accumulated/run.go
@@ -27,7 +27,6 @@ import (
 	"github.com/tendermint/tendermint/crypto"
 	tmlog "github.com/tendermint/tendermint/libs/log"
 	tmnet "github.com/tendermint/tendermint/libs/net"
-	tmrand "github.com/tendermint/tendermint/libs/rand"
 	"github.com/tendermint/tendermint/privval"
 	rpchttp "github.com/tendermint/tendermint/rpc/client/http"
 	"github.com/tendermint/tendermint/rpc/client/local"
@@ -126,14 +125,10 @@ func (d *Daemon) Start() (err error) {
 	testDir := d.Config.BaseConfig.RootDir
 	sockPath := filepath.Join(testDir, "test.sock")
 
-	tmpSock := "unix:///tmp/accum." + tmrand.Str(6) + ".sock"
-
 	d.Config.RPC.ListenAddress = fmt.Sprintf("unix://%s", sockPath)
-	d.Config.Accumulate.API.ListenAddress = tmpSock
 	d.Listen = fmt.Sprintf("unix://%s", sockPath)
 	d.Protocol = "unix"
 
-	defer os.Remove(tmpSock)
 	defer os.Remove(sockPath)
 
 	dbPath := filepath.Join(d.Config.RootDir, "valacc.db")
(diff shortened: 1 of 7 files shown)


firelizzard18 (changes requested):
The PR description must include a description of the changes made and instructions on how to verify the changes. See [CONTRIBUTING.md](https://github.com/AccumulateNetwork/accumulate/blob/develop/CONTRIBUTING.md).

internal/accumulated/run.go
```diff
@@ -342,6 +360,19 @@ func listenHttpUrl(s string) (net.Listener, bool, error) {
...
+func listenTCPorUnix(s string) (net.Listener, error) {
+	protocol, address := tmnet.ProtocolAndAddress(s)
+	l, err := net.Listen(protocol, address)
+	if err != nil {
+		return nil, err
+	}
+
+	return l, nil
+}
+
```
firelizzard18, line 375: Instead of creating a new function, modify the existing one. `listenHttpUrl` is still an appropriate name, because it's still listening on HTTP, and we should still leave room for HTTPS. Instead of using `tmnet.ProtocolAndAddress`, add a case in `listenHttpUrl` for `u.Scheme == "unix"` and skip the path check if the scheme is unix.

internal/testing/node.go
```diff
@@ -37,24 +38,37 @@ func DefaultConfig(net config.NetworkType, node config.NodeType, netId string) *
 func NodeInitOptsForNetwork(subnet *networks.Subnet) node.InitOptions {
 	listenIP := make([]string, len(subnet.Nodes))
 	remoteIP := make([]string, len(subnet.Nodes))
+	var listener string
 	cfg := make([]*config.Config, len(subnet.Nodes))
 
 	for i, net := range subnet.Nodes {
+		listener = "unix:///tmp/accumulate.sock"
```
firelizzard18, line 45:
```suggestion
		listener = fmt.Sprintf("unix:///tmp/accumulate-%d.sock", i)
```

internal/testing/node.go
```diff
@@ -37,24 +38,37 @@ func DefaultConfig(net config.NetworkType, node config.NodeType, netId string) *
...
+	var la *net.UnixAddr
+	la = &net.UnixAddr{
+		Name: listener,
+		Net:  "unix",
+	}
+	ks, err := net.ListenUnix("unix", la)
+	if err != nil {
+		panic(err)
+	}
+	ks.SyscallConn()
```
firelizzard18, line 59: What does this do?

firelizzard18 (changes requested):
I'm expecting one socket per port. Tendermint has three services, p2p, rpc, rpc-grpc, and we have one service, the api (if Prometheus is disabled, which it pretty much always has to be for tests). So I'm expecting the tests will have to open four unix sockets.

internal/testing/node.go
```diff
@@ -37,24 +38,37 @@ func DefaultConfig(net config.NetworkType, node config.NodeType, netId string) *
...
 	}
+	var la *net.UnixAddr
+	la = &net.UnixAddr{
+		Name: listener,
+		Net:  "unix",
+	}
+	ks, err := net.ListenUnix("unix", la)
+	if err != nil {
+		panic(err)
+	}
```
firelizzard18, line 58: Why are we starting to listen here? All this function does is generate configuration.

PR #331, opened 2021-12-09
Ac 689 get any state

PaulSnow:
GetAnyState provides the ability to retrieve from the database the state of a Merkle Tree at any element, despite the fact that we only save states into the database at the beginning of a Mark Set.

We do this by finding the previous mark point to an element, feed the entries into that MerkleState up to the element.

This is super handy in building Merkle Proofs.

In the course of implementing GetAnyState, I found a number of difficult to track down bugs, some bad implementation, and fixes broke tests which I repaired.

Receipt tests are now broken, but I wanted to do those fixes in the next pull request.


firelizzard18 (changes requested):

smt/managed/merklemanager.go
```diff
@@ -211,27 +213,79 @@ func (m *MerkleManager) init(DBManager *database.Manager, markPower int64) (err
...
-	if element == 0 {
-		return new(MerkleState)
+	if m.GetElementCount() == 0 {
+		ms := new(MerkleState)
+		if eHash, err := m.Get(element); err != nil {
+			ms.AddToMerkleTree(eHash)
+		}
+		ms.InitSha256()
+		return ms
 	}
```
firelizzard18, line 231: What does this do?

types/state/chain_manager.go
```diff
@@ -35,7 +35,7 @@ func (c *ChainManager) Height() int64 {
 
 // Record returns the record.
 func (c *ChainManager) Record() (*Object, error) {
-	data, err := c.state.dbMgr.Key(append(c.key, "Record")...).Get()
+	data, err := c.state.dbMgr.Key(append(c.key, "Record")).Get()
```
firelizzard18, line 38: `storage.ComputeKey` does not handle `[]interface{}`, so this will panic, I believe.

types/state/chain_manager.go
```diff
@@ -132,7 +135,7 @@ func (c *ChainManager) Update(record []byte) error {
 	h := sha256.Sum256(data)
 	c.state.logDebug("Updating chain state", "key", c.key, "hash", logging.AsHex(h))
 
-	c.state.dbMgr.Key(append(c.key, "Record")...).PutBatch(data)
+	c.state.dbMgr.Key(c.key, "Record").PutBatch(data)
```
firelizzard18, line 138: I expect this will panic

types/state/chain_manager.go
```diff
@@ -73,7 +73,10 @@ func (c *ChainManager) Entries(start int64, end int64) ([][]byte, error) {
 	// multiple times
 	entries := make([][]byte, 0, end-start)
 	for start < end {
-		h, err := c.merkle.GetRange(c.key, start, end)
+		if err := c.merkle.SetKey(c.key...); err != nil {
+			return nil, err
+		}
```
firelizzard18, line 78: `ChainManager` holds its own copy of the merkle manager with the key set, so this is unnecessary.

smt/managed/merklemanager.go
```diff
@@ -76,30 +76,30 @@ func (m *MerkleManager) GetElementIndex(hash []byte) (i int64, err error) {
...
 		return err
 	}
-	m.Manager.Key(append(m.key, "Head")...).PutBatch(state)
+	m.Manager.Key(append(key, "Head")...).PutBatch(state)
 	return nil
 }
 
 // GetChainState
-// Reads the highest state of the chain stored to the database.  Returns nil
+// Reads the highest state of the chain stored to the database.  Returns nilTestCli
```
firelizzard18, line 97:
```suggestion
// Reads the highest state of the chain stored to the database.  Returns nil
```

firelizzard18 (approved):

PR #332, opened 2021-12-09
Ac 617 write data tx

bunfield:
- implemented WriteData validator for chain entries
- implemented a segwit approach to decouple transactions from data payload to prevent data from being stored twice (once with the transactions and once with the data chain).
- added computation of the WriteData cost and capped data @ 10kB / transaction (and subsequently stubbed out a fee schedule that can be leveraged in a later task) 
- enabled nonce in GenTransaction SignatureInfo because a test was failing because it was setting default nonce to 1
- deferred all query functionality to the next task (API updates for data) so unit tests will be expanded in the next task


internal/abci/utils_test.go
```diff
@@ -163,6 +163,11 @@ func (n *fakeNode) GetChainStateByUrl(url string) *api.APIDataResponse {
 	return r
 }
 
+func (n *fakeNode) GetChainDataByUrl(url string) *api.APIDataResponse {
+	//todo implement query data functionality
+	return nil
+}
```
firelizzard18, line 169: I would prefer a panic or `n.t.Fatal(...)`, to make it immediately obvious why the test failed

internal/abci/utils_test.go
```diff
@@ -284,6 +289,11 @@ type e2eDUT struct {
 	*fakeNode
 }
 
+func (n e2eDUT) GetDataByUrl(url string) (*ctypes.ResultABCIQuery, error) {
+	//todo: implement query data by url
+	return nil, nil
+}
```
firelizzard18, line 295: Same comment about panic/Fatal

internal/chain/state.go
```diff
@@ -212,6 +220,20 @@ func (m *StateManager) UpdateNonce(record state.Chain) {
 	m.store(record, updateNonce)
 }
 
+//UpdateCreditBalance update the credits used for a transaction
+func (m *StateManager) UpdateCreditBalance(record state.Chain) {
+	//todo implement
+}
```
firelizzard18, line 226: I would prefer a panic to `//todo implement`

firelizzard18 (changes requested):

firelizzard18 (approved):

firelizzard18:
We've been getting failures on macOS and Windows builders, but normally only on one job at a time

bunfield:
running the test on my windurs laptop i get a failure at

bunfield:
So it appears the problem is after a send(tx) in the e2e tests you need to add a sleep so the channel doesn't get closed before the tx is finished being processes.  It may be the source of other seemingly random failures on Windows and Mac.

PR #333, opened 2021-12-09
Clean up config types

firelizzard18:
The big change is that `config.Accumulated.Networks` and `.Directory` are going away, and instead we will have a list of BVN names and a map from DN/BVN name to address. This particular change is necessary for DN to BVN mirroring and anchoring.

- Eliminate `RPC`
- Eliminate `API.RESTListenAddress`
- Rename `API.{JSON,}ListenAddress`
- Move website-related config into its own type
- Move network type, id into its own type
- Replace `Accumulated.Networks` and `.Directory` with `Network.BvnNames` and `.Addresses`
- Add helpers in config for changing the port number

And update everything touched by these changes.

No functional changes. To verify, run the node like normal and test it.


PaulSnow (changes requested):
Reviewed, but I don't have as much experience with this part of the code.
@bunfield Should review.

bunfield (approved):

bunfield:
Looks good to me.

PaulSnow (approved):

PR #335, opened 2021-12-10
Ac 677 key exist check

ranjan008:
API to get key indices from key page or keybook


internal/chain/executor_query.go
```diff
@@ -223,6 +224,84 @@ func (m *Executor) Query(q *query.Query) (k, v []byte, err *protocol.Error) {
...
+			return nil, nil, &protocol.Error{Code: protocol.CodeInvalidURL, Message: fmt.Errorf("invalid URL in query %s", chr.Url)}
+		}
+		obj, err := m.queryByChainId(u.ResourceChain())
+		if err != nil {
+			return nil, nil, &protocol.Error{Code: protocol.CodeAddKeyPageIndices, Message: err}
+		}
+		chainHeader := new(state.ChainHeader)
+		if err = obj.As(chainHeader); err != nil {
+			return nil, nil, &protocol.Error{Code: protocol.CodeMarshallingError, Message: fmt.Errorf("inavid object error")}
+		}
```
firelizzard18, line 244: If the record is something other than a key book (or page?), then we need to check the record's key book.

```suggestion
		}
		if chainHeader.Type != types.ChainTypeKeyBook {
			obj, err = m.queryByChainId(chainHeader.KeyBook)
			if err != nil {
				return nil, nil, &protocol.Error{Code: protocol.CodeAddKeyPageIndices, Message: err}
			}
			err = obj.As(chainHeader)
			if err = obj.As(chainHeader); err != nil {
				return nil, nil, &protocol.Error{Code: protocol.CodeMarshallingError, Message: fmt.Errorf("inavid object error")}
			}
		}
```

protocol/errors.go
```diff
@@ -46,6 +46,8 @@ const (
 	CodeInvalidTxnError ErrorCode = 22
 	//CodeAddTxnError is returned when adding txn to state db fails
 	CodeAddTxnError ErrorCode = 23
+	//CodeAddKeyPageIndices is returned when fetching key indices fail
+	CodeAddKeyPageIndices ErrorCode = 24
```
firelizzard18, line 50: Why is this "Add"? Regardless, I think I would prefer returning `CodeNotFound` if the key is not found.

types/api/APIRequests.go
```diff
@@ -48,6 +48,11 @@ type APIRequestURL struct {
 	Wait bool         `json:"wait" form:"wait" query:"wait"`
 }
 
+type APIRequestKeyPageIndices struct {
+	APIRequestURL
+	Key string `json:"key" form:"key" query:"key" validate:"required"`
+}
+
```
firelizzard18, line 55: This should be removed. As far as I can tell, it isn't used anywhere, and regardless, types for API v2 should be added to `internal/api/v2/types.yml`.

types/api/query/chain.go
```diff
@@ -12,13 +13,36 @@ type RequestByUrl struct {
...
+
+type KeyPageResponse struct {
+	KeyPage string `json:"key_page"`
+	Indices []int  `json:"indices"`
+}
+
+type ResponseKeyPageIndices struct {
+	KeyBook        string             `json:"key_book"`
+	KeyPageIndices []*KeyPageResponse `json:"key_page_indices"`
+}
```
firelizzard18, line 29: This needs to be moved to `internal/api/v2/types.yml`. Types specific to API v2 do not belong here.

firelizzard18 (changes requested):

firelizzard18 (changes requested):
Everything looks fine, but the API call does not find my key: `curl -X POST --data '{ "jsonrpc": "2.0", "id": 0, "method": "query-key-index", "params": { "url": "keytest", "key": 
"'${PUBLIC_KEY}'" } }' -H 'Content-Type: application/json' http://127.0.1.1:26660/v2`

firelizzard18 (approved):

PR #343, opened 2021-12-13
Use docker compose for deploy and validation

firelizzard18:
- Now you can use docker compose to run a network
- Now CI runs the validation script _before_ merge, using docker compose to deploy a testnet for the CI job

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+.nodes
+*.exe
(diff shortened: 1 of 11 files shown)


bunfield (approved):
Other than the ci validation failing. Nothing stood out as being wrong.  @ilzheev may be a beer candidate for review as well since he has more docker experience.

PR #345, opened 2021-12-15
feat(sdk): migrated cli func to apiV2

brohamgoham:


libs/goaccumulate/goaccumulate.go
```diff
@@ -0,0 +1,247 @@
...
+	"github.com/AccumulateNetwork/accumulate/cmd/cli/db"
+	api2 "github.com/AccumulateNetwork/accumulate/internal/api/v2"
+	url2 "github.com/AccumulateNetwork/accumulate/internal/url"
+	"github.com/AccumulateNetwork/accumulate/protocol"
+	"github.com/mdp/qrterminal"
+)
+
+var (
+	Client         = client.NewAPIClient()
+	Db             db.DB
```
bunfield, line 22: The database should not be in sdk. key management is handled outside the sdk.

libs/goaccumulate/goaccumulate.go
```diff
@@ -0,0 +1,247 @@
...
+	TxPretend      = false
+)
+
+var (
+	BucketAnon     = []byte("anon")
+	BucketAdi      = []byte("adi")
+	BucketKeys     = []byte("keys")
+	BucketLabel    = []byte("label")
+	BucketMnemonic = []byte("mnemonic")
+)
```
bunfield, line 33: no bucket definitions should be here

libs/goaccumulate/goaccumulate.go
```diff
@@ -0,0 +1,247 @@
...
+		WhiteChar:      qrterminal.WHITE_WHITE,
+		WhiteBlackChar: qrterminal.WHITE_BLACK,
+		QuietZone:      2,
+	})
+
+	r, err := ioutil.ReadAll(b)
+	return string(r), err
+}
+
+func CreateAccount(url string, args []string) (string, error) {
```
bunfield, line 115: the arguments passed to this function should not be an array of strings.  should be either the data structure of explicit parameters

libs/goaccumulate/goaccumulate.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+	binaryData, err := tac.MarshalBinary()
+	if err != nil {
+		return "", err
+	}
+
+	nonce := nonceFromTimeNow()
+	params, err := prepareGenTxV2(jsonData, binaryData, actor, si, privKey, nonce)
+	if err != nil {
```
bunfield, line 174: we will not always have access to the private key.  For example, when passing on to a hardware wallet for signing.  need a signing callback or something similar

libs/goaccumulate/goaccumulate.go
```diff
@@ -0,0 +1,247 @@
...
+		return "", err
+	}
+
+	var res api2.TxResponse
+
+	if err := Client.RequestV2(context.Background(), "create-token-account", params, &res); err != nil {
+		return PrintJsonRpcError(err)
+	}
+
+	return ActionResponseFrom(&res).Print()
```
bunfield, line 184: the printing is customized for the cli and should not be part of the sdk.

libs/goaccumulate/goaccumulate.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+func GenerateAccount() (string, error) {
+	return GenerateKey("")
+}
+
+func ListAccounts() (string, error) {
+
```
bunfield, line 192: this function doesn't belong in the sdk.

libs/goaccumulate/goaccumulate.go
bunfield: perhaps explore the use of a builder pattern instead.

libs/goaccumulate/goaccumulate.go
```diff
@@ -0,0 +1,247 @@
...
+		}
+		if lt.String() == string(v.Key) {
+			out += fmt.Sprintf("%s\n", v.Key)
+		}
+	}
+	return out, nil
+}
+
+func RestoreAccounts() (out string, err error) {
+	anon, err := Db.GetBucket(BucketAnon)
```
bunfield, line 212: does not belong...

libs/goaccumulate/key.go
```diff
@@ -0,0 +1,396 @@
+package goaccumulate
+
```
bunfield, line 2: key management should be handled by the wallet not the sdk.

libs/goaccumulate/page.go
```diff
@@ -0,0 +1,210 @@
...
+		return "", fmt.Errorf("error retrieving key page for %s, %v", url, err)
+	}
+
+	res := api2.QueryResponse{}
+
+	err = json.Unmarshal(str, &res)
+	if err != nil {
+		return "", err
+	}
+	return PrintQueryResponseV2(&res)
```
bunfield, line 25: the return needs to be basic type or just json. the print is custom to the cli

libs/goaccumulate/page.go
```diff
@@ -0,0 +1,210 @@
...
+	if err := Client.RequestV2(context.Background(), "create-key-page", params, &res); err != nil {
+		return PrintJsonRpcError(err)
+	}
+
+	ar := ActionResponse{}
+	//	err = json.Unmarshal(*res.Data, &ar)
+	//	if err != nil {
+	//		return "", fmt.Errorf("error unmarshalling create adi result, %v", err)
+	//	}
+	return ar.Print()
```
bunfield, line 118: see comments above about returning prints

bunfield (changes requested):
instead of putting this in the libs directory, it should be under pkg (https://github.com/golang-standards/project-layout).  however, given this will require a lot more work, we should probably put it in internal/pkg until api is settled, at which time we can make it public.

I made comments to the first couple of files, but the comments carry throughout the code and apply to all files in this PR.

firelizzard18:
Moving to GitLab

PR #346, opened 2021-12-15
Fix init

firelizzard18:
One of my PRs introduced a bug in init that broke the devnet deployment


bunfield (approved):
First time in a few days all checks passed.. :+1:

PR #354, opened 2021-12-17
CI test reports

firelizzard18:
Use `gotest.tools/gotestsum` to run tests and generate a JUnit test report file. This will improve the readability of test results in CI, and GitLab CI can use the JUnit report to populate a test results UI.


bunfield (approved):
ran it and it worked as advertised.

PR #358, opened 2021-12-21
Remove old transaction types

firelizzard18:
Move SendTokens and SyntheticDepositTokens to protocol (via types.yml), remove redundant fields on those two typers, and remove types/synthetic.


bunfield (changes requested):
Starting at line 252 in cmd/cli/cmd/tx.go it is referencing the token send in api2..  this is a bug that is fixed in a later version that is now OBE by this PR.  I would suggest fixing that bug with this PR instead.

bunfield:
> Starting at line 252 in cmd/cli/cmd/tx.go it is referencing the token send in api2.. this is a bug that is fixed in a later version that is now OBE by this PR. I would suggest fixing that bug with this PR instead.

Here is the offending code (it is using the wrong tokensend).  It somehow passed the automated tests, but it causes the dispatchTx to fail.
`	tokenSend := new(api2.TokenSend)
	tokenSend.From = u.String()

	amt, err := strconv.ParseFloat(amount, 64)
	r := api2.TokenDeposit{}
	r.Amount = uint64(amt * 1e8)
	r.Url = u2.String()
	tokenSend.To = append(tokenSend.To, r)
`

bunfield (approved):

PR #359, opened 2021-12-21
Fix token send

bunfield:
- The orig issue was to fix cli error but that was done as part of PR #358, however, this PR did have a useful fix to the human-readable output, so the PR merged the fix from develop and only includes cleanup of output and v1 lingo
- cleaned up some human-readable output on tx get
- removed legacy RequestV1/V2 for the client and replaced it with client.Request


bunfield:
This PR is OBE by https://github.com/AccumulateNetwork/accumulate/pull/358, but leaving it for the moment.

firelizzard18 (approved):

PR #360, opened 2021-12-21
Ac 743 api router error

ranjan008:
Test case to send token transaction to different BVN's where origin is not a valid account


config/config.go
```diff
@@ -31,7 +31,8 @@ const (
 	Follower  NodeType = "follower"
 )
 
-const DefaultLogLevels = "error;main=info;state=info;statesync=info;accumulate=debug;executor=info"
+// const DefaultLogLevels = "error;main=info;state=info;statesync=info;accumulate=debug;executor=info"
+const DefaultLogLevels = "error"
```
firelizzard18, line 35: Please revert this - I forgot about it. This was a change I made just to make the test log cleaner.

firelizzard18 (changes requested):

firelizzard18 (approved):

PR #371, opened 2021-12-24
fix: CLI docker image

firelizzard18:


firelizzard18:
Pushing this through to fix CI

PR #375, opened 2021-12-27
AC 786 remove tx reference

ranjan008:


firelizzard18 (changes requested):
We also need to remove:
- `types.ChainTypeTransactionReference` ([here](https://github.com/AccumulateNetwork/accumulate/blob/1b9bc0ae4867d9a604ef7b09c30d59b9c4d0054a/types/chain_types.go#L34))
- `state.TxReference` ([here](https://github.com/AccumulateNetwork/accumulate/blob/b19dbb4af5d047c212ef386cd4fc238bc7ec0712/types/state/TxReference.go#L11))
- Code in the validators for SendTokens and SyntheticDepositTokens that creates TxReference objects ([here](https://github.com/AccumulateNetwork/accumulate/blob/81f58bec31ec680ac17e6ee59cd9e08d215fa19f/internal/chain/send_tokens.go#L74-L79) and [here](https://github.com/AccumulateNetwork/accumulate/blob/81f58bec31ec680ac17e6ee59cd9e08d215fa19f/internal/chain/synthetic_deposit_tokens.go#L66-L71))

firelizzard18 (approved):

